PR #58, opened 2024-03-26
Update package to 0.4.1

GGonryun:
Bumps the minor version for our latest release which includes a breaking change for eventually supporting making an access request for a group of nodes.


komal-dhull (approved):
lgtm

PR #60, opened 2024-03-26
Users can start a session for nodes with a group approval

GGonryun:
This PR allows users to connect to nodes that have been granted group approval.

### DIFF
--- a/src/commands/ssh.ts
+++ b/src/commands/ssh.ts
@@ -163,7 +163,7 @@ const ssh = async (args: yargs.ArgumentsCamelCase<SshCommandArgs>) => {
   // Prefix is required because the backend uses it to determine that this is an AWS request
   const authn = await authenticate();
   await validateSshInstall(authn);
-  const response = await request(
+  const response = await request<AwsSsh>(
     {
       ...pick(args, "$0", "_"),
       arguments: [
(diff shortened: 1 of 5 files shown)


src/types/request.ts
```diff
@@ -31,10 +31,11 @@ export type Request<P extends PluginRequest = { permission: object }> = {
   principal: string;
 };
 
-export type RequestResponse = {
+export type RequestResponse<T> = {
   ok: true;
   message: string;
   id: string;
+  event: T;
```
GGonryun, line 38: The `event` payload has always been returned. We're just exposing it now via an interface.

src/commands/ssh.ts
```diff
@@ -184,11 +184,11 @@ const ssh = async (args: yargs.ArgumentsCamelCase<SshCommandArgs>) => {
     print2("Did not receive access ID from server");
     return;
   }
-  const { id, isPreexisting } = response;
+  const { id, isPreexisting, event } = response;
   if (!isPreexisting) print2("Waiting for access to be provisioned");
 
   const requestData = await waitForProvisioning<AwsSsh>(authn, id);
-  const requestWithId = { ...requestData, id };
+  const requestWithId = { ...requestData, id, permission: event.permission };
```
GGonryun, line 191: We rely on the returned permission object instead of waiting for the results of `waitForProvisioning` because SSH group requests workflow documents differ from SSH session request workflow documents. 

If we use the permission value from the SSH group request document we won't have a real instance arn to ssh into.

gergas3 (approved):
lgtm

PR #61, opened 2024-03-26
Update package to 0.4.2

GGonryun:
This PR updates the `package.json` to version 0.4.2 for our latest release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@p0security/cli",
-  "version": "0.4.1",
+  "version": "0.4.2",
   "description": "Execute infra CLI commands with P0 grants",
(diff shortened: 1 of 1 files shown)


nbrahms (approved):

nbrahms (approved):
Just to check, this is not already published, correct?

We do need to make sure version changes are merged _before_ we publish. Otherwise we can't point to the code version that backs an artifact.

GGonryun:
> Just to check, this is not already published, correct?
> 
> We do need to make sure version changes are merged _before_ we publish. Otherwise we can't point to the code version that backs an artifact.

Not published yet, I made this PR preemptively 👍

PR #63, opened 2024-03-28
Change ssh instance ls display to use instance ids instead of full arn

komal-dhull:
The other info in the ARN is redundant since it's already available from the "group"

image

### DIFF
--- a/src/commands/ls.ts
+++ b/src/commands/ls.ts
@@ -12,7 +12,7 @@ import { fetchCommand } from "../drivers/api";
 import { authenticate } from "../drivers/auth";
 import { guard } from "../drivers/firestore";
 import { print2, print1, Ansi } from "../drivers/stdio";
-import { max, isEqual} from "lodash";
+import { max, isEqual } from "lodash";
 import pluralize from "pluralize";
 import yargs from "yargs";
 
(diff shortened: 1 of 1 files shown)


src/commands/ls.ts
```diff
@@ -41,6 +41,14 @@ export const lsCommand = (yargs: yargs.Argv) =>
     guard(ls)
   );
 
+
+/** Applies any command-specific formatting to the p0 ls display */
+const displayLsResource = (args: string[], key: string) =>
```
nbrahms, line 46: 🤔 we should avoid anything resource-specific in our base / core code. We should instead separate concerns. I think maybe a better way to do this is to add something like `lsKey` to the `ListerOption` type instead, and generically use that.

komal-dhull: That makes sense, and I see your point. I did consider adding it to the `ListerOption`, but I was concerned it didn't make sense to have this logic live in the API? Since this is purely a display concern where the data returned from the API is sufficient and the CLI just wants to manipulate it prior to the display, I would have expected it to be in the CLI (and also because we don't want this to apply to the Slack slash commands). Do you think that's a concern? (I'm happy to implement it that way, just looking to better understand the reasoning)

PR #64, opened 2024-03-29
aws: Update for new install data model

nbrahms:
P0's data model for AWS installations has changed. Update the CLI to work with this new data model.

### DIFF
--- a/src/commands/__tests__/ssh.test.ts
+++ b/src/commands/__tests__/ssh.test.ts
@@ -32,9 +32,9 @@ mockGetDoc({
   workflows: {
     items: [
       {
-        identifier: "test-account",
         state: "installed",
         type: "aws",
+        identifier: "test-account",
       },
     ],
   },
--- a/src/commands/aws/role.ts
+++ b/src/commands/aws/role.ts
@@ -13,7 +13,7 @@ import { authenticate } from "../../drivers/auth";
 import { guard } from "../../drivers/firestore";
 import { print1, print2 } from "../../drivers/stdio";
 import { getAwsConfig } from "../../plugins/aws/config";
-import { AwsItemConfig, AwsOktaSamlUidLocation } from "../../plugins/aws/types";
+import { AwsFederatedLogin, AwsItem } from "../../plugins/aws/types";
 import { assumeRoleWithOktaSaml } from "../../plugins/okta/aws";
 import { getSamlResponse } from "../../plugins/okta/login";
 import { Authn } from "../../types/identity";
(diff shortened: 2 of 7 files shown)


src/commands/__tests__/ssh.test.ts
```diff
@@ -32,9 +32,9 @@ mockGetDoc({
   workflows: {
     items: [
       {
-        identifier: "test-account",
```
GGonryun, line 35: 🤔

src/commands/aws/role.ts
```diff
@@ -46,10 +46,10 @@ export const role = (yargs: yargs.Argv<{ account: string | undefined }>) =>
...
   );
 
-const isOktaSamlConfig = (
-  config: AwsItemConfig
-): config is AwsItemConfig & { uidLocation: AwsOktaSamlUidLocation } =>
-  config.uidLocation?.id === "okta_saml_sso";
+const isFederatedLogin = (
+  config: AwsItem
+): config is AwsItem & { login: AwsFederatedLogin } =>
+  config.login?.type === "federated";
```
gergas3, line 52: Should this be the equivalent stricter check that verifies all the way down to `login.provider.method == saml` (and `okta`, etc.)?

src/plugins/aws/config.ts
```diff
@@ -23,11 +24,19 @@ export const getAwsConfig = async (
...
+    ),
+    ([id]) => id
+  );
   const item = account
-    ? config?.workflows?.items.find(
-        (i) => i.state === "installed" && i.account.id === account
-      )
-    : config?.workflows?.items[0];
+    ? allItems.find(([id, { label }]) => id === account || label === account)
+    : allItems.length !== 1
```
gergas3, line 35: Can this ever be more than 1? `account` or `label` would have to match multiple items.
Suggestion: just pick the first one like we used to.

gergas3 (approved):

GGonryun (approved):
👍

src/commands/aws/role.ts
nbrahms: not necessary I think; when we add more complications we can update this

src/plugins/aws/config.ts
nbrahms: it doesn't make any sense to just pick the first one though; I discovered this by trying to `p0 aws role ls`, which would randomly fail or pass depending on wether the first was the one that was federated login

src/commands/__tests__/ssh.test.ts
nbrahms: yeah, debugging artifact 🤷

PR #68, opened 2024-04-03
Add support for requesting groups during ssh request

GGonryun:
This PR introduces a new argument to our cli: `p0 ssh --request-group private-node`. If this flag is added the cli will first make a group permission request before attempting to ssh into the requested node. This argument does not work with sudo yet since sudo is not supported when requesting access to a group of nodes.

This depends on changes to the  `p0/app` repository: <PLACE TICKET HERE>.


src/commands/ssh.ts
```diff
@@ -163,6 +169,27 @@ const ssh = async (args: yargs.ArgumentsCamelCase<SshCommandArgs>) => {
...
   await validateSshInstall(authn);
+
+  if (args.requestGroup) {
+    // When requesting for a group we need 2 requests. The first request grants access to the group
+    // requests will detect that the node is a part of the group and can use the group permission
+    // to gain access to the node.
+    const response = await request(
+      {
+        ...pick(args, "$0", "_"),
+        arguments: ["ssh", "group", "--destination", args.destination],
```
nbrahms, line 180: how does this work with pre-existing access?

GGonryun: There's a different bug for groups right now where making a request to a group with pre-existing access creates a duplicate request. I made a separate  PR to fix it independently here: https://github.com/p0-security/app/pull/1569

Without the fix we'll keep requesting new access requests for a group of nodes. With the fix we'll be able to gain access without making another request.

Essentially these two scenarios will work without generating any extra requests:
1. `p0 request ssh group Miguel` and then `p0 ssh --request-group private-node`
2. `p0 ssh --request-group private-node` and `p0 ssh --request-group private-node`

gergas3 (commented):
What is the use case for the `request-group` sub-command in a `p0 ssh`?

I get the idea of `p0 request group --destination <destination>`: you know one node and want to extend your access to the entire group by _requesting_ that access you don't have.

However, the `p0 ssh` command is the intent to land inside a node. Why does the user think about "requesting" and "landing" at the same time? Or rather, why does the user think about the "how"? It's too specific to p0. Originally we wanted `p0 ssh` look like regular ssh as much as possible.

So I find this command confusing at first glance. Based on my current understanding I think we should not add it.

nbrahms (approved):
cool, thanks for the explanation

GGonryun:
> What is the use case for the `request-group` sub-command in a `p0 ssh`?
> 
> I get the idea of `p0 request group --destination <destination>`: you know one node and want to extend your access to the entire group by _requesting_ that access you don't have.
> 
> However, the `p0 ssh` command is the intent to land inside a node. Why does the user think about "requesting" and "landing" at the same time? Or rather, why does the user think about the "how"? It's too specific to p0. Originally we wanted `p0 ssh` look like regular ssh as much as possible.
> 
> So I find this command confusing at first glance. Based on my current understanding I think we should not add it.

Even the basic command `p0 ssh private-node` does both the "requesting" and "landing" since it essentially does `p0 request ssh session private-node` and then `p0 ssh private-node`. With `--request-group` we change up the model so that instead of doing a session request, we create a group request. So it's not that much different than what's currently happening.

I am lean slightly towards not adding it because it keeps the ssh command simple and and users can still perform the request independently using `p0 request ssh group --destination <destination>`. @nbrahms, what do you think?

nbrahms:
> > What is the use case for the `request-group` sub-command in a `p0 ssh`?
> > I get the idea of `p0 request group --destination <destination>`: you know one node and want to extend your access to the entire group by _requesting_ that access you don't have.
> > However, the `p0 ssh` command is the intent to land inside a node. Why does the user think about "requesting" and "landing" at the same time? Or rather, why does the user think about the "how"? It's too specific to p0. Originally we wanted `p0 ssh` look like regular ssh as much as possible.
> > So I find this command confusing at first glance. Based on my current understanding I think we should not add it.
> 
> Even the basic command `p0 ssh private-node` does both the "requesting" and "landing" since it essentially does `p0 request ssh session private-node` and then `p0 ssh private-node`. With `--request-group` we change up the model so that instead of doing a session request, we create a group request. So it's not that much different than what's currently happening.
> 
> I am lean slightly towards not adding it because it keeps the ssh command simple and and users can still perform the request independently using `p0 request ssh group --destination <destination>`. @nbrahms, what do you think?

The point here is that in the `p0 ssh` model, we anticipate most users asking for approval by running `p0 ssh` directly, rather than distinctly making access requests. If this is true, users will need to repeatedly make new approval requests for each node.

We could wait to see if this is a problem, though.

gergas3:
> > > What is the use case for the `request-group` sub-command in a `p0 ssh`?
> > > ...
> > > So I find this command confusing at first glance. Based on my current understanding I think we should not add it.
> > 
> > 
> > Even the basic command `p0 ssh private-node` does both the "requesting" and "landing" since it essentially does `p0 request ssh session private-node` and then `p0 ssh private-node`. With `--request-group` we change up the model so that instead of doing a session request, we create a group request. So it's not that much different than what's currently happening.
> > I am lean slightly towards not adding it because it keeps the ssh command simple and and users can still perform the request independently using `p0 request ssh group --destination <destination>`. @nbrahms, what do you think?
> 
> The point here is that in the `p0 ssh` model, we anticipate most users asking for approval by running `p0 ssh` directly, rather than distinctly making access requests. If this is true, users will need to repeatedly make new approval requests for each node.
> 
> The high-level question: should users ever need to know about `p0 request`?
> 
> We could wait to see if this is a problem, though.

I think users probably don't want to know about how access is given to a node if they enter a specific node name. If they want to request a group by specifying a node in the group there is the `p0 request` command.

gergas3:
> should users ever need to know about p0 request?

They don't need to know about it until they want to tune the "how" part of the request. 
I think it's confusing to include more request options other than the ones needed to identify the requested object.

GGonryun:
> I think users probably don't want to know about how access is given to a node if they enter a specific node name. If they want to request a group by specifying a node in the group there is the `p0 request` command.

Would we be okay with punting on this change until more users are using `p0 ssh`. Customers that are asking for this feature provide a clearer signal of demand versus the opposite: customers that aren't using the feature even though it's there.

gergas3:
Yeah

> > I think users probably don't want to know about how access is given to a node if they enter a specific node name. If they want to request a group by specifying a node in the group there is the `p0 request` command.
> 
> Would we be okay with punting on this change until more users are using `p0 ssh`. Customers that are asking for this feature provide a clearer signal of demand versus the opposite: customers that aren't using the feature even though it's there.

Yes, that's fair I think

PR #72, opened 2024-04-10
Set AWS_SECURITY_TOKEN environment variable

gergas3:
`AWS_SECURITY_TOKEN` is the legacy version of `AWS_SESSION_TOKEN`. 

It does seem to take precedence over `AWS_SESSION_TOKEN`. 

The okta-aws-cli sets both: https://github.com/okta/okta-aws-cli/blob/f1e09eab509e295a7e7b3002d14f2a96b8c60914/internal/output/envvar.go#L49L63

This causes an issue where if someone uses the okta-aws-cli to log in and then uses p0, the environment variables end up in  an inconsistent state.


GGonryun (approved):
LGTM! Squash that bug! ❌ 🐛

PR #73, opened 2024-04-18
SCP command for copying from local to remote

gergas3:
Adds `p0 scp <source> <destination>` command to copy a single file from local to remote.

The instance name can be used (while the instance IP cannot):

https://github.com/p0-security/p0cli/assets/5836460/27f47de5-5ce0-4c9a-bb25-5cd95632be6b

We detect if the source / destination args refer to a local or a remote file using the same rules as the real `scp` command, however copying from remote to local is not supported and throws an error.

The implementation uses two ssm subprocesses:
1. Open a port using the AWS-StartPortForwardingSession document
2. Run a non-interactive command `nc -l -p ${port} > ${destination}` on the remote - this waits for the data to arrive on the remote and dumps it into a file

In the local machine a third operation writes the file contents to the specified port. (The same port is used on the local and the remote.) This is done using a TCP socket from the `node:net`package instead of running a local subprocess with netcat.
Reading the file is synchronous and the transfer is not chunked. That will be optimized in a follow-up.

Other things I tried:

- Using a pipe and reading from stdin on the remote
      ```cat configmap.yml.zip | aws ssm start-session --region us-west-2 --target i-043db199de9ac0c53 --document P0SshAsUser-FYEbC0gpZpvigekLuGMe-gergely-danyi --parameters command='cat - > configmap.yml.zip'```
     Unfortunately this doesn't work, likely an issue with how ssm is implemented.
- Real SCP by running a AWS-StartPortForwardingSession document to port 22 of the remote (with a running OpenSSH server). This works but requires a keypair and we do not want to manage keys.

The PR moves some functions for reuse with scp.

### DIFF
--- a/src/commands/ssh.ts
+++ b/src/commands/ssh.ts
@@ -8,43 +8,14 @@ This file is part of @p0security/cli
 
 You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
 **/
-import { authenticate } from "../drivers/auth";
-import { doc, guard } from "../drivers/firestore";
-import { print2 } from "../drivers/stdio";
-import { ssm } from "../plugins/aws/ssm";
-import { AwsSsh } from "../plugins/aws/types";
-import { SshConfig } from "../plugins/ssh/types";
-import { Authn } from "../types/identity";
-import {
-  DENIED_STATUSES,
-  DONE_STATUSES,
-  ERROR_STATUSES,
-  PluginRequest,
-  Request,
-} from "../types/request";
-import { request } from "./request";
-import { getDoc, onSnapshot } from "firebase/firestore";
-import { pick } from "lodash";
+import { guard } from "../drivers/firestore";
+import { sshRequest } from "../plugins/aws/ssm/request";
+import { ssmSsh } from "../plugins/aws/ssm/ssh";
+import { SshCommandArgs } from "./types";
 import yargs from "yargs";
-
-export type SshCommandArgs = {
-  destination: string;
-  command?: string;
-  L?: string; // Port forwarding option
-  N?: boolean; // No remote command
-  arguments: string[];
-  sudo?: boolean;
-  reason?: string;
-  account?: string;
-};
 
 // Matches strings with the pattern "digits:digits" (e.g. 1234:5678)
 const LOCAL_PORT_FORWARD_PATTERN = /^\d+:\d+$/;
-
-/** Maximum amount of time to wait after access is approved to wait for access
- *  to be configured
- */
-const GRANT_TIMEOUT_MILLIS = 60e3;
 
 export const sshCommand = (yargs: yargs.Argv) =>
   yargs.command<SshCommandArgs>(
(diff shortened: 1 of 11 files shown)


src/commands/ssh.ts
```diff
@@ -8,44 +8,17 @@ This file is part of @p0security/cli
...
-export type SshCommandArgs = {
-  destination: string;
-  command?: string;
-  L?: string; // Port forwarding option
-  N?: boolean; // No remote command
-  arguments: string[];
-  sudo?: boolean;
-  reason?: string;
-  account?: string;
-};
```
gergas3, line 39: Moved to `types.ts`

src/commands/ssh.ts
```diff
@@ -8,44 +8,17 @@ This file is part of @p0security/cli
...
 
 // Matches strings with the pattern "digits:digits" (e.g. 1234:5678)
 const LOCAL_PORT_FORWARD_PATTERN = /^\d+:\d+$/;
 
-/** Maximum amount of time to wait after access is approved to wait for access
- *  to be configured
- */
-const GRANT_TIMEOUT_MILLIS = 60e3;
-
```
gergas3, line 48: Moved to `request.ts`

src/commands/ssh.ts
```diff
@@ -101,62 +74,6 @@ export const sshCommand = (yargs: yargs.Argv) =>
...
-  );
-  const configItems = configDoc.data()?.["iam-write"];
-
-  const items = Object.entries(configItems ?? {}).filter(
-    ([key, value]) => value.state == "installed" && key.startsWith("aws")
-  );
-  if (items.length === 0) {
-    throw "This organization is not configured for SSH access via the P0 CLI";
-  }
-};
```
gergas3, line 116: Moved to `request.ts`

src/commands/ssh.ts
```diff
@@ -101,62 +74,6 @@ export const sshCommand = (yargs: yargs.Argv) =>
...
-      if (!isResolved) {
-        unsubscribe();
-        reject("Timeout awaiting SSH access grant");
-      }
-    }, GRANT_TIMEOUT_MILLIS);
-  });
-  clearTimeout(cancel);
-  return result;
-};
-
```
gergas3, line 159: Moved to `request.ts`

src/commands/ssh.ts
```diff
@@ -165,48 +82,11 @@ const waitForProvisioning = async <P extends PluginRequest>(
  * - AWS EC2 via SSM with Okta SAML
  */
 const ssh = async (args: yargs.ArgumentsCamelCase<SshCommandArgs>) => {
```
gergas3, line 84: The code below moved to `request.ts#sshRequest`

src/plugins/aws/ssm/index.ts
```diff
@@ -148,29 +147,11 @@ const createPortForwardingCommand = (
...
-  });
-
-  if (args.noRemoteCommands) {
-    return { shellCommand: portForwardingCommand };
-  }
-
-  return {
-    shellCommand: interactiveShellCommand,
-    subCommand: portForwardingCommand,
-  };
```
gergas3, line 173: Moved to `ssh.ts`.
This `index.ts` file now contains functions used by both ssh and scp.

src/plugins/aws/ssm/request.ts
```diff
@@ -0,0 +1,136 @@
+/** Copyright © 2024-present P0 Security
```
gergas3, line 1: This file contains only moved functions

src/plugins/aws/ssm/ssh.ts
```diff
@@ -0,0 +1,76 @@
+/** Copyright © 2024-present P0 Security
```
gergas3, line 1: This file contains only moved functions from `index.ts`

gergas3:
Closing this because we need full compatibility with SCP

PR #74, opened 2024-04-23
Add support for SCP and exercising grants

GGonryun:
## Overview

This PR adds support for SCP and the new `exercise-grant` endpoint.

This is PR 2/2
1. [p0-security/app #1623](https://github.com/p0-security/app/pull/1623)
2. [p0-security/p0cli #74](https://github.com/p0-security/p0cli/pull/74) 👈

`p0 scp` supports:
1. Uploading files from local host to remote host.
2. Download files from remote host to local host.
4. File transfers up to 2GB.
5. Downloading or Uploading the entire contents of a folder (using `-r` argument)
6. 10 concurrent file transfer operations.

## Examples
image

### Upload File
`p0 scp 10mb.txt private-node:100mb.txt -i mykey.pem`

### Download File
`p0 scp private-node:100mb.txt 100mb.txt -i mykey.pem`

### Sudo Access
`p0 scp 100mb.txt private-node:100mb.txt -i mykey.pem --sudo`

## Future Work
See future work planned on Linear: https://linear.app/p0-security/issue/ENG-1911/scp-support-with-public-key-transfer#comment-6a36eb05

## Manual Test Plan
See manual test plan on Linear: https://linear.app/p0-security/issue/ENG-1911/scp-support-with-public-key-transfer#comment-85500292


gergas3 (approved):
I tested it in my dev environment, working well!
I only had minor comments.

src/commands/scp.ts
```diff
@@ -0,0 +1,151 @@
...
+    ...args,
+    source,
+    destination,
+  });
+};
+
+/** If a path is not explicitly local, use this pattern to determine if it's remote */
+const REMOTE_PATTERN_COLON = /^([^:]+:)(.*)$/; // Matches host:[path]
+
+const isExplicitlyRemote = (path: string): boolean => {
```
gergas3, line 117: One thing to note is that detecting the remote / local paths is not complete. SCP has more complex logic around that. E.g. there could be colons in the local path which you can specify by starting path with `.`, `..`, or `/`

GGonryun: Good point, I created a ticket for expanding on the more complex logic for SCP local and remote paths.

https://linear.app/p0-security/issue/ENG-1931/improve-remote-host-and-local-host-checking-for-scp-requests

src/commands/scp.ts
```diff
@@ -0,0 +1,151 @@
...
+  const isSourceRemote = isExplicitlyRemote(source);
+  const isDestinationRemote = isExplicitlyRemote(destination);
+
+  const remote = isSourceRemote ? source : destination;
+
+  if (isSourceRemote != isDestinationRemote) {
+    return remote.split(":")[0];
+  }
+
+  throw "At least one host (source or destination) must be remote.";
```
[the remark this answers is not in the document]
GGonryun: Good point! I created a todo to support multiple remote hosts. I added a comment above this error referencing the ticket.

https://linear.app/p0-security/issue/ENG-1930/support-scp-across-multiple-external-hosts

src/plugins/aws/ssm/index.ts
```diff
@@ -246,12 +262,40 @@ const spawnSsmNode = async (
...
+ * or logout/login cycles. Once you log out or restart your system, any keys added to the SSH agent during that session
+ * will need to be added again in subsequent sessions.
+ */
+const executeScpCommand = async (
+  credential: AwsCredentials,
+  command: string,
+  privateKey: string
+) => {
+  // Execute should not leave any ssh-agent processes running after it's done performing an SCP transaction.
+  const execute = `
```
GGonryun, line 283: Things started to get really complex when I tried to run all these commands as separate processes. I would have needed to find a way to share environment variables across sub-processes, make sure that tty was working for scp, and then do the cleanup after SCP was done (more abort controller shenanigans). Instead it's easier to just execute the commands in the order we need them.

src/plugins/aws/ssm/index.ts
```diff
@@ -266,6 +268,34 @@ const spawnSsmNode = async (
...
+  eval $(ssh-agent) >/dev/null 2>&1
+  trap 'kill $SSH_AGENT_PID' EXIT
+  ssh-add -q - <<< '${privateKey}'
```
gergas3, line 286: Why is `<<<` important?

src/plugins/aws/ssm/index.ts
```diff
@@ -266,6 +268,34 @@ const spawnSsmNode = async (
...
+ */
+const executeScpCommand = async (
+  credential: AwsCredentials,
+  command: string,
+  privateKey: string
+) => {
+  // Execute should not leave any ssh-agent processes running after it's done performing an SCP transaction.
+  const execute = `
+  eval $(ssh-agent) >/dev/null 2>&1
+  trap 'kill $SSH_AGENT_PID' EXIT
```
gergas3, line 285: I'm not sure we have to take on the job of running and killing the ssh agent. Most of the time it will already be running, and in that case the user will not expect that p0 stops it.

We could check if it's running and ask the user to run it if it's not. I tried and if I rerun `eval (ssh-agent)` when it's already running it will spawn a new process, presumably with empty ssh-key state.

Then we could add and remove by trapping `ssh-add -d ~/.ssh/sshkeynamewithout.pub`? Not sure that would work.

gergas3 (commented):
Looking good! The only concern is storing and transmitting private keys comes with some risk we have to think about. Is it possible to generate the key pair in the client?

For the PoC we could generate the keypair in the CLI, and ensure that the SSH agent has a p0-generated key that we can use to make the connection. Imo it's OK for this key to survive sessions and even access requests.

In the longer term we will probably use client certificates like Okta ASA does:
https://www.okta.com/blog/2020/07/ssh-is-dead-long-live-ssh-one-million-ssh-logins-with-okta-zero-ssh-keys/

gergas3 (approved):

src/plugins/aws/ssm/index.ts
```diff
@@ -445,44 +477,40 @@ const createScpCommand = (
   return [
     "scp",
     "-o",
-    `ProxyCommand=${ssmCommand.join(" ")}`,
+    `ProxyCommand='${ssmCommand.join(" ")}'`,
     // if a response is not received after three 5 minute attempts,
     // the connection will be closed.
     "-o",
+    "ForwardAgent=yes",
```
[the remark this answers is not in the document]
GGonryun: It supports the later but we don't really need this right now so I removed it. I was experimenting with SCP across remote destinations and the third-party transfers (where a third host can act as an intermediary for the file transfer)

src/plugins/aws/ssm/index.ts
GGonryun: It redirects the string as input. otherwise ssh-add interprets it as an option or a path incorrectly. Which is weird to me because we're still escaping things with a single quote.

> ssh-add: unrecognized option `-----BEGIN RSA PRIVATE KEY-----\r

src/plugins/aws/ssm/index.ts
GGonryun: > I'm not sure we have to take on the job of running and killing the ssh agent. Most of the time it will already be running, and in that case the user will not expect that p0 stops it.

I mostly did this to prevent the `ssh-agent` DOS and to prevent sharing keys with other running agents.

If the user has a bunch of other keys in the `ssh-agent` it will try to use each key to make a connection request. I found that having 5+ unrelated keys blocks me out of the host because it'll think I'm trying to break by attempting to connect many times in a split second frequently.

We won't kill any other running agents, only the ssh-agent we spawn get's killed, running `eval $(ssh-agent)` spawns a new agent and since we're running as a child process we shouldn't mess with any other process' environment  variables

PR #76, opened 2024-04-29
Do not allow private key to appear in bash history or process logs

GGonryun:
This PR introduces a new method of executing our SCP command. Our previous method accidentally included the ssh key in the name which accidentally revealed the private key. This new method starts a `bash` process and then pipes in the `scp` command through `stdin` preventing the private key from being revealed. We're also unsetting that subcommands history, this should prevent anyone who somehow gains access to that subcommand from searching for the private key in the command history.

## Before The Fix
image

## After The Fix
image

### DIFF
--- a/src/plugins/aws/ssm/index.ts
+++ b/src/plugins/aws/ssm/index.ts
@@ -507,13 +507,14 @@ export const scp = async (
   });
 
   /**
-   * Spawns a child process to add a private key to the ssh-agent. The SSH agent is included in the OpenSSH suite of tools
-   * and is used to hold private keys during a session. The SSH agent typically does not persist keys across system reboots
-   * or logout/login cycles. Once you log out or restart your system, any keys added to the SSH agent during that session
-   * will need to be added again in subsequent sessions.
+   * Spawns a child process to add a private key to the ssh-agent. The SSH agent is included in the OpenSSH suite
+   * of tools and is used to hold private keys during a session. The SSH agent typically does not persist keys
+   * across system reboots or logout/login cycles. Once you log out or restart your system, any keys added to
+   * the SSH agent during that session will need to be added again in subsequent sessions.
    */
   const writeStdin = [
-    // completely disable bash history for this session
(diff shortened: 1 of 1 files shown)


src/plugins/aws/ssm/index.ts
```diff
@@ -280,19 +327,20 @@ const executeScpCommand = async (
...
-  ssh-add -q - <<< '${privateKey}' 
-  ${command}
-  SCP_EXIT_CODE=$? 
-  exit $SCP_EXIT_CODE
-  `;
-
-  return spawnSsmNode({
+  const commands = [
+    // completely disable bash history for this session
+    `unset HISTFILE`,
```
GGonryun, line 332: This _might_ be overkill because we're spawning a subprocess that will run the commands for us but just in case somehow someone enters that subprocess we're also disabling the history.

src/plugins/aws/ssm/index.ts
```diff
@@ -500,16 +497,38 @@ export const scp = async (
...
-    throw "Failed to load a private key for this request. Please contact support@p0.dev for assistance.";
-  }
+  /**
+   * Spawns a child process to add a private key to the ssh-agent. The SSH agent is included in the OpenSSH suite
+   * of tools and is used to hold private keys during a session. The SSH agent typically does not persist keys
+   * across system reboots or logout/login cycles. Once you log out or restart your system, any keys added to
+   * the SSH agent during that session will need to be added again in subsequent sessions.
+   */
+  const writeStdin = [
+    // This_might be overkill because we are already spawning a subprocess that will run the commands for us
```
gergas3, line 516: typo
```suggestion
    // This might be overkill because we are already spawning a subprocess that will run the commands for us
```

gergas3 (approved):
lgtm

src/plugins/aws/ssm/index.ts
GGonryun: fixed 👍

PR #77, opened 2024-04-30
Add debug mode for scp

GGonryun:
This PR enables a debug mode for SCP that:
- displays information about the `ssh-agent`
- prints the private key that will be used to connect to the requested node.
- provides manual commands to run for local testing.

### Example
**Sample Command**
```
p0 scp 10mb.txt clone-node-2:10mb.txt --debug
```

image

### DIFF
--- a/src/plugins/aws/ssm/index.ts
+++ b/src/plugins/aws/ssm/index.ts
@@ -553,22 +553,22 @@ export const scp = async (
     `exit $SCP_EXIT_CODE`,
   ];
 
-  // Print commands that can be individually executed to reproduce behavior
-  // Remove the debug information - can be executed manually between steps
-  const reproCommands = without(
-    [
-      "bash",
-      ...Object.entries(process.env).map(
-        ([key, value]) => `export ${key}='${value}'`
-      ),
-      ...Object.entries(credential).map(
-        ([key, value]) => `export ${key}='${value}'`
-      ),
-      ...writeStdin,
-    ],
-    ...debug
-  );
   if (args.debug) {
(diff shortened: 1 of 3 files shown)


src/plugins/aws/ssm/index.ts
```diff
@@ -516,10 +542,13 @@ export const scp = async (
...
+    `eval $(ssh-agent)${args.debug ? "" : " >/dev/null 2>&1"}`,
     `trap 'kill $SSH_AGENT_PID' EXIT`,
     `ssh-add -q - <<< '${privateKey}'`,
-    createScpCommand(data, args).join(" "),
+    // in debug mode, we'll see the keys that were added to the agent and more information about the agent
+    ...(args.debug ? debug : []),
+    scpCommand,
     `SCP_EXIT_CODE=$?`,
     `exit $SCP_EXIT_CODE`,
   ];
```
gergas3, line 554: Suggestion: also print the sequence of commands that can be used to create a similar SCP flow. I tested this and seems to work.

```suggestion
  ];

  // Print commands that can be individually executed to reproduce behavior
  // Remove the debug information - can be executed manually between steps
  const reproCommands = without([
    "bash",
    ...Object.entries(process.env).map(([key, value]) => `export ${key}='${value}'`),
    ...Object.entries(credential).map(([key, value]) => `export ${key}='${value}'`),
    ...writeStdin
  ], ...debug);
  print2(`Execute the following commands to create a similar SCP session:\n *** COMMANDS BEGIN ***\n${reproCommands.join("\n")}\n *** COMMANDS END ***`);
```

gergas3 (approved):
👍

PR #80, opened 2024-05-16
Ping Identity OIDC login

gergas3:
Adds support for logging in with Ping Identity (Tested with PingOne)

The flow from Okta can be entirely reused.
The first commit shows the actual code changes by editing `src/plugins/okta/login.ts` in place.
The second commit is a pure code move to extract the shared functions + addition of a missed copyright header.
(Third commit is code formatting.)

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -85,3 +85,15 @@ export const exec = async (
       }
     }
   );
+
+
+export const throwAssertNever = (value: never) => {
+  throw assertNever(value);
+};
+
+export const assertNever = (value: never) => {
+  return unexpectedValueError(value);
+};
+
+export const unexpectedValueError = (value: any) =>
+  new Error(`Unexpected code state: value ${value} had unexpected type`);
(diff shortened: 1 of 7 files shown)


src/util.ts
```diff
@@ -85,3 +85,14 @@ export const exec = async (
       }
     }
   );
+
+export const throwAssertNever = (value: never) => {
```
GGonryun, line 89: small nit but I don't think this is a significant advantage over `throw assertNever(value)` we save on typing whitespace and this potentially messes with the stack trace since the throw would come from this file and not the original location.

GGonryun (approved):
left one tiny nit. i'm okay with this being merged as is 👍

src/util.ts
gergas3: The issue I had is that `throw assertNever(value)` doesn't compile in a ternary expression:
image

GGonryun: I see, TIL!

PR #81, opened 2024-06-11
Agent-forwarding in ssh command with -A and ssh_config's ForwardAgent option

gergas3:
Add support for forwarding the ssh-agent in the `p0 ssh` command

- Change the SCP code path to use `shell=true` instead of passing `bash` as command and writing to the stdin of the subprocess
- This allows reusing the same subprocess spawning method for SSH: if there is no non-interactive command and no port-forwarding in the `p0 ssh` command then use the `AWS-StartSshSession` document like the SCP path
- The `-o "ForwardAgent=yes"` option is not supported
- Replace the `bash` subprocess with separate subprocesses that do not require a shell
- Do not print sensitive data (the private key) in debug mode. Instead keep the ssh-agent alive so it can be used with manual execution of the repro commands

In these videos `ssh` is aliased to call `p0 ssh`:

https://github.com/p0-security/p0cli/assets/5836460/75a5c285-9ff8-40b2-9b07-5393a3de3225

https://github.com/p0-security/p0cli/assets/5836460/11b61b63-a20a-4b4f-8dda-ce9525af2822

### DIFF
--- a/proxy-ssh-private-node.sh
+++ b/proxy-ssh-private-node.sh
@@ -1,1 +1,0 @@
-ssh -i /Users/gergo/Documents/certificate-based-ssh/gergely.danyi@p0.dev -t -t -o "CertificateFile=/Users/gergo/Documents/certificate-based-ssh/gergely.danyi@p0.dev-cert.pub" -o ProxyCommand='aws ssm start-session --region us-west-2 --target %h --parameters "portNumber=%p" --document-name AWS-StartSSHSession' $@
--- a/proxy-ssh-private-node.sh
+++ b/proxy-ssh-private-node.sh
@@ -1,0 +1,1 @@
+ssh -i /Users/gergo/Documents/certificate-based-ssh/gergely.danyi@p0.dev -t -t -o "CertificateFile=/Users/gergo/Documents/certificate-based-ssh/gergely.danyi@p0.dev-cert.pub" -o ProxyCommand='aws ssm start-session --region us-west-2 --target %h --parameters "portNumber=%p" --document-name AWS-StartSSHSession' $@
--- a/src/plugins/aws/ssm/index.ts
+++ b/src/plugins/aws/ssm/index.ts
@@ -507,7 +507,7 @@ const createProxyCommands = (
     ...commonArgs,
     ...(args.A ? ["-A"] : []),
     `${data.linuxUserName}@${data.instance.id}`,
-  ];
+  ].join(" ");
 };
 
 export const sshOrScp = async (
@@ -529,7 +529,7 @@ export const sshOrScp = async (
     role: data.role,
   });
 
-  const command = createProxyCommands(data, args, args.debug).join(" ");
+  const command = createProxyCommands(data, args, args.debug);
 
   const debug = [
     `echo "SSH_AUTH_SOCK: $SSH_AUTH_SOCK"`,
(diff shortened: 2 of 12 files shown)


proxy-ssh-private-node.sh
GGonryun, line 1: Should these be checked in? 🤔

src/plugins/aws/ssm/index.ts
```diff
@@ -521,16 +529,13 @@ export const scp = async (
     role: data.role,
   });
 
-  const commands = createProxyCommands(data, args, args.debug);
-  const scpCommand = commands.scp.join(" ");
-  const sshCommand = commands.ssh.join(" ");
+  const command = createProxyCommands(data, args, args.debug).join(" ");
```
GGonryun, line 532: nit: we could just return the joined command, do we ever use it as an array?

proxy-ssh-private-node.sh
gergas3: No...

src/plugins/aws/ssm/index.ts
```diff
@@ -225,19 +228,21 @@ function spawnChildProcess(
     env: {
       ...process.env,
       ...credential,
+      ...(sshAgentEnv || {}),
```
GGonryun, line 231: I thought using the spread operator `...` on an undefined value just did nothing 🤔

src/plugins/ssh-agent/index.ts
```diff
@@ -0,0 +1,146 @@
...
+) =>
+  new Promise<number>((resolve, reject) => {
+    const child = spawn(command, args, options);
+
+    if (writeStdin) {
+      if (!child.stdin) return reject("Child process has no stdin");
+      child.stdin.write(writeStdin);
+    }
+
+    child.stdout.on("data", (data) => {
```
GGonryun, line 37: DRY: could we share the callback function somehow? 🤔

src/plugins/ssh-agent/index.ts
```diff
@@ -0,0 +1,146 @@
...
+      child.stdin?.end();
+    }
+  });
+
+/** Spawns a subprocess with the ssh-agent command.
+ * Detects the auth socket and agent PID from stdout.
+ * Stdout and stderr of the subprocess is printed to stderr in debug mode.
+ * The returned promise resolves with an object that contains the auth socket and agent PID,
+ * or rejects with the contents of stderr. */
+export const sshAgent = async (cmdArgs: AgentArgs) =>
```
GGonryun, line 66: Do we spawn a new ssh-agent? or are we sharing the existing ssh-agent? If so, are we sure that doing something like `ctrl-c` terminates the agent process as well?

GGonryun (approved):
LGTM in general, just got a few questions.

src/plugins/aws/ssm/index.ts
gergas3: It compiles here but that's surprising to me... 
I tried with a toy example:
image

src/plugins/ssh-agent/index.ts
gergas3: Good point! 
`ctrl-d` aka closing the session terminates the agent, however if `ctrl-c` is pressed after the agent is running but before the session is started then the agent remained alive.
I addressed that by listening to termination signals from the main process and killing the agent process in the listener.

src/plugins/ssh-agent/index.ts
gergas3: I tried but I think it would require "mutating" a string inside a function and the mutation should be reflected outside. I would probably need an object or class to do so but then it becomes too much code imo.

GGonryun: understandable 👍

PR #86, opened 2024-07-09
Use persistent keys for granting ssh access

GGonryun:
This PR removes the ephemeral keys and exercise-grant endpoint. Instead a pair of keys will be generated and stored in the `.p0` folder. These keys will be used every time you access a node. These keys are automatically inserted into a users running agent. If an agent is not running, they will receive an error message telling them to start one.

image

New SSH debug commands:

image

#### Sudo node access
1. Perform a regular ssh access request: `p0 ssh private-node --sudo`
#### SCP node access
1. Perform a request with SCP using: `p0 scp private-node:10mb.txt test.txt`
#### Agent Forwarding
1. Add a few keys to your agent using `ssh-add`.
2. Ssh into target machine with agent forwarding argument: `p0 ssh private-node -A`
3. Run `ssh-add -l` from target machine and verify that ssh-agent was forwarded.
#### Multiple Sockets
1. Run `pgrep -x ssh-agent` to find your ssh agents.
2. Run `eval $(ssh-agent)` to spawn an extra ssh agent.
3. Run `lsof -p $pid` to find the socket for an agent.
4. Set the socket using `SSH_AUTH_SOCK=$socket`
5. Execute `p0 ssh $instance --debug` with debug flag. 
    1. The new ssh agent should insert a key
    2. The old ssh agent should skip inserting a key.
6. Change the socket:
    1. SSH_AUTH_SOCK=/private/tmp/com.apple.launchd.qDjtwp32tR/Listeners ./p0 ssh private-node --debug
##### Teardown
1. Connect to a node using `p0 ssh`.
2. Relinquish your connection.
3. Log into the node using SSM on the AWS GUI.
4. Verify that the user's authorized key file (/home/miguel-campos/.ssh/authorized_keys) is empty
#### Non-Ephemeral Access
1. Connect to a node using `p0 ssh`.
2. Verify that your authorized keys has 1 key.
3. Disconnect from the node.
4. Connect to the same node again.
5. Verify that your authorized keys still has 1 key.

### DIFF
--- a/src/commands/scp.ts
+++ b/src/commands/scp.ts
@@ -8,15 +8,15 @@ This file is part of @p0security/cli
 
 You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
 **/
-import { fetchExerciseGrant } from "../drivers/api";
 import { authenticate } from "../drivers/auth";
 import { guard } from "../drivers/firestore";
 import { sshOrScp } from "../plugins/aws/ssm";
 import {
-  ExerciseGrantResponse,
   ScpCommandArgs,
+  SshRequest,
   createKeyPair,
   provisionRequest,
+  requestToSsh,
 } from "./shared";
 import yargs from "yargs";
 
@@ -75,26 +75,22 @@ const scpAction = async (args: yargs.ArgumentsCamelCase<ScpCommandArgs>) => {
     throw "Could not determine host identifier from source or destination";
   }
 
-  const requestId = await provisionRequest(authn, args, host);
+  const { publicKey, privateKey } = await createKeyPair();
 
-  if (!requestId) {
+  const request = await provisionRequest(authn, args, host, publicKey);
+
+  if (!request) {
     throw "Server did not return a request id. Please contact support@p0.dev for assistance.";
   }
 
-  const { publicKey, privateKey } = createKeyPair();
-
-  const result = await fetchExerciseGrant(authn, {
-    requestId,
-    destination: host,
-    publicKey,
-  });
+  const data = requestToSsh(request);
 
   // replace the host with the linuxUserName@instanceId
-  const { source, destination } = replaceHostWithInstance(result, args);
+  const { source, destination } = replaceHostWithInstance(data, args);
 
   await sshOrScp(
     authn,
-    result,
+    data,
     {
       ...args,
       source,
@@ -127,19 +123,16 @@ const getHostIdentifier = (source: string, destination: string) => {
   throw "Exactly one host (source or destination) must be remote.";
 };
 
-const replaceHostWithInstance = (
-  result: ExerciseGrantResponse,
-  args: ScpCommandArgs
-) => {
+const replaceHostWithInstance = (result: SshRequest, args: ScpCommandArgs) => {
   let source = args.source;
   let destination = args.destination;
 
   if (isExplicitlyRemote(source)) {
-    source = `${result.linuxUserName}@${result.instance.id}:${source.split(":")[1]}`;
+    source = `${result.linuxUserName}@${result.id}:${source.split(":")[1]}`;
   }
 
   if (isExplicitlyRemote(destination)) {
-    destination = `${result.linuxUserName}@${result.instance.id}:${destination.split(":")[1]}`;
+    destination = `${result.linuxUserName}@${result.id}:${destination.split(":")[1]}`;
   }
 
   return { source, destination };
--- a/src/commands/scp.ts
+++ b/src/commands/scp.ts
@@ -8,13 +8,13 @@ This file is part of @p0security/cli
 
 You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
 **/
+import { createKeyPair } from "../common/keys";
 import { authenticate } from "../drivers/auth";
 import { guard } from "../drivers/firestore";
 import { sshOrScp } from "../plugins/aws/ssm";
 import {
   ScpCommandArgs,
   SshRequest,
-  createKeyPair,
   provisionRequest,
   requestToSsh,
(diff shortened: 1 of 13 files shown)


src/common/keys.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (
+    (await fileExists(PUBLIC_KEY_PATH)) &&
+    (await fileExists(PRIVATE_KEY_PATH))
+  ) {
+    const publicKey = await fs.readFile(PUBLIC_KEY_PATH, "utf8");
+    const privateKey = await fs.readFile(PRIVATE_KEY_PATH, "utf8");
+
+    return { publicKey, privateKey };
+  } else {
+    const rsaKeyPair = forge.pki.rsa.generateKeyPair({ bits: 2048 });
```
p0-andrewa, line 35: More a question/discussion point rather than something that I think should be changed in this PR: Was there a previous decision regarding selecting 2048-bit RSA keys for SSH? Ed25519 is a more modern standard for SSH keys which produces much smaller keys (so considerably easier to deal with than RSA keys for, say, passing in as a CLI argument or pasting into Slack) but still has security comparable to (if not perhaps a bit better than) 3072-bit RSA keys. It even has better performance, but wouldn't be noticeable for humans really. The only drawback is that Ed25519 is a newer standard so some very old SSH servers might not accept them, but anything reasonably modern should support it (OpenSSH added support for Ed25519 keys with version 6.5 in January 2014).

p0-andrewa (commented):
LGTM but will defer on final approval. Did have a question inline about the choice of SSH key algorithm, though.

src/commands/ssh.ts
```diff
@@ -85,22 +85,16 @@ const sshAction = async (args: yargs.ArgumentsCamelCase<SshCommandArgs>) => {
 
   const destination = args.destination;
 
-  const requestId = await provisionRequest(authn, args, destination);
-  if (!requestId) {
+  const { publicKey, privateKey } = await createKeyPair();
```
gergas3, line 88: nit: `createKeyPair` could move inside `provisionRequest` here and in scp.ts

src/plugins/ssh-agent/index.ts
```diff
@@ -59,100 +56,71 @@ const asyncSpawn = async (
...
-      if (!authSockMatch?.[1] || !agentPidMatch?.[1]) {
-        return reject("Failed to parse ssh-agent stdout:\n" + stdout);
-      }
-      resolve({
-        SSH_AUTH_SOCK: authSockMatch[1],
-        SSH_AGENT_PID: agentPidMatch[1],
-      });
-    });
-  });
+const isSshAgentAuthSocketSet = async (args: AgentArgs) => {
```
gergas3, line 71: Is the auth socket env var enough? Running ssh-agent also sets `SSH_AGENT_PID`, isn't that needed?

src/plugins/ssh-agent/index.ts
```diff
@@ -59,100 +56,71 @@ const asyncSpawn = async (
...
-      if (code !== 0) {
-        return reject(stderr);
-      }
-
-      const authSockMatch = stdout.match(AUTH_SOCK_MESSAGE);
-      const agentPidMatch = stdout.match(AGENT_PID_MESSAGE);
+const isSshAgentRunning = async (args: AgentArgs) => {
+  try {
+    if (args.debug) print2("Searching for active ssh-agents");
+    await asyncSpawn(args, `pgrep`, ["-x", "ssh-agent"]);
```
gergas3, line 62: What if you have ssh-agent process(es) but your `SSH_AGENT_PID` is set to an invalid process ID that either doesn't exist or a different type of process. Would the actual ssh command try to use that?
Add a TODO if you think this is a valid edge case?

src/commands/scp.ts
```diff
@@ -75,26 +75,22 @@ const scpAction = async (args: yargs.ArgumentsCamelCase<ScpCommandArgs>) => {
     throw "Could not determine host identifier from source or destination";
   }
 
-  const requestId = await provisionRequest(authn, args, host);
+  const { publicKey, privateKey } = await createKeyPair();
 
-  if (!requestId) {
+  const request = await provisionRequest(authn, args, host, publicKey);
```
gergas3, line 80: If the request is pre-existing then the returned request object from Datastore must have the same publicKey as the one we submitted from `createKeyPair`. That's going to be the case as long as our backend implementation is correct but we could also add a check here that compares the two public keys. wdyt?

gergas3 (approved):

src/plugins/ssh-agent/index.ts
GGonryun: Yeah it turns out that we don't need a PID, as long as the auth socket is set everything works fine. 👍

src/plugins/ssh-agent/index.ts
GGonryun: TLDR: I'll write a TODO here. It's not an impossible edge case, but improbable.

Technically we don't use `SSH_AGENT_PID` for anything but we're just checking to make sure an ssh-agent is running. In theory its possible that:

1. someone starts an ssh-agent.
2. someone starts a new ssh-agent without changing the auth socket.
3. someone kills the old ssh agent.

but when they try to gain ssh access they'll get a permission denied error. if they run with a `--debug` flag it'll be very obvious that they're running the wrong agent/socket though. 

It wouldn't be impossible for us tho but it would take some `shell` magic to make sure that the PID + socket match and that they are both active.

src/common/keys.ts
GGonryun: We were using RSA's for now for the ephemeral generation, and in theory it also works for Ed25519, but since we're planning to move towards certificates the choice of algorithm has been deferred.

src/commands/scp.ts
GGonryun: 👍 , good idea. i added a small check here.

PR #87, opened 2024-07-11
Send a body request when assuming roles with SAML

GGonryun:
This PR changes the request for assuming roles from using `GET` to `POST`. 

Our SAML Assertion will be too long for the encoded url search params if we are assuming too many roles. Instead we can send a `POST` message with the request in the body to get around length requirements.


nbrahms:
Note that package size is not equivalent to install size. `npm -g install` installs dependencies separately; they are not bundled in `dist`.

But I think size would matter less compared to potential distribution issues (I'm not sure if things like Python are required to install the `aws-sdk`, we'd want to investigate that).

nbrahms (approved):
lgtm

PR #88, opened 2024-07-12
Release v0.8.0

GGonryun:
Release v0.8.0 which moves away from ephemeral keys and exercise grants, it also fixes a bug where assuming roles with saml failed if too many roles were requested.


gergas3 (approved):
Why not 0.8.0?

GGonryun:
> Why not 0.8.0?

good point, arrays start at 0 🤔

PR #90, opened 2024-07-22
Basic gcloud ssh

gergas3:
Support ssh with gcloud using native ssh with public key authentication and ProxyCommand - the same way as the existing AWS ssh works.

The underlying ProxyCommand in gcloud's case is `gcloud compute start-iap-tunnel`. The IAP tunnel is necessary for accessing VMs without an external IP.

All of the existing SSH functionality is supported: `-A`, `-N`, `-L`, `--sudo`, and `scp`.

However, login to Google Cloud is separate from `p0 login`. There is no support for automatic log in to Google Cloud with SAML (or other way). If we detect that the issue is gcloud login we print `Please login to Google Cloud CLI with 'gcloud auth login'`

- Rename variables and rewrite comments to apply to both AWS and GCP
- Rework types to more closely resemble backend types and share common fields
- Add detection gcloud-specific access propagation detection

### DIFF
--- a/src/commands/scp.ts
+++ b/src/commands/scp.ts
@@ -10,7 +10,7 @@ You should have received a copy of the GNU General Public License along with @p0
 **/
 import { authenticate } from "../drivers/auth";
 import { guard } from "../drivers/firestore";
-import { sshOrScp } from "../plugins/aws/ssm";
+import { sshOrScp } from "../plugins/ssh";
 import {
   ScpCommandArgs,
   provisionRequest,
--- a/src/commands/scp.ts
+++ b/src/commands/scp.ts
@@ -84,9 +84,9 @@ const scpAction = async (args: yargs.ArgumentsCamelCase<ScpCommandArgs>) => {
   if (!result) {
     throw "Server did not return a request id. Please contact support@p0.dev for assistance.";
   }
-  const { request, privateKey } = result;
+  const { request, privateKey, linuxUserName } = result;
 
-  const data = requestToSsh(request);
+  const data = requestToSsh(request, { gcloud: { linuxUserName } });
 
   // replace the host with the linuxUserName@instanceId
   const { source, destination } = replaceHostWithInstance(data, args);
--- a/src/commands/scp.ts
+++ b/src/commands/scp.ts
@@ -13,9 +13,9 @@ import { guard } from "../drivers/firestore";
 import { sshOrScp } from "../plugins/aws/ssm";
 import {
   ScpCommandArgs,
-  SshRequest,
   provisionRequest,
   requestToSsh,
+  SshRequest,
 } from "./shared";
 import yargs from "yargs";
 
@@ -48,6 +48,11 @@ export const scpCommand = (yargs: yargs.Argv) =>
         .option("account", {
           type: "string",
           describe: "The account on which the instance is located",
+        })
+        .option("provider", {
+          type: "string",
+          describe: "The cloud provider where the instance is hosted",
+          choices: ["aws", "gcloud"],
         })
         .option("sudo", {
           type: "boolean",
--- a/src/commands/scp.ts
+++ b/src/commands/scp.ts
@@ -84,9 +84,9 @@ const scpAction = async (args: yargs.ArgumentsCamelCase<ScpCommandArgs>) => {
   if (!result) {
     throw "Server did not return a request id. Please contact support@p0.dev for assistance.";
   }
-  const { request, privateKey, linuxUserName } = result;
+  const { request, privateKey } = result;
 
-  const data = requestToSsh(request, { gcloud: { linuxUserName } });
+  const data = requestToSsh(request);
 
   // replace the host with the linuxUserName@instanceId
   const { source, destination } = replaceHostWithInstance(data, args);
--- a/src/commands/scp.ts
+++ b/src/commands/scp.ts
@@ -16,6 +16,7 @@ import {
   provisionRequest,
   requestToSsh,
   SshRequest,
+  SUPPORTED_PROVIDERS,
 } from "./shared";
 import yargs from "yargs";
 
@@ -52,7 +53,7 @@ export const scpCommand = (yargs: yargs.Argv) =>
         .option("provider", {
           type: "string",
           describe: "The cloud provider where the instance is hosted",
-          choices: ["aws", "gcloud"],
+          choices: SUPPORTED_PROVIDERS,
         })
         .option("sudo", {
           type: "boolean",
--- a/src/commands/shared.ts
+++ b/src/commands/shared.ts
@@ -147,13 +147,17 @@ const waitForProvisioning = async <P extends PluginRequest>(
 };
 
 const pluginToCliRequest = async (
-  request: Request<PluginRequest>
+  request: Request<PluginRequest>,
+  options?: { debug?: boolean }
 ): Promise<Request<CliRequest>> => {
   return request.permission.spec.type === "gcloud"
     ? ({
         ...request,
         cliLocalData: {
-          linuxUserName: await importSshKey(request.permission.spec.publicKey),
+          linuxUserName: await importSshKey(
+            request.permission.spec.publicKey,
+            options
+          ),
         },
       } as Request<GcpSsh>)
     : request.permission.spec.type === "aws"
@@ -205,7 +209,9 @@ export const provisionRequest = async (
     throw "Public key mismatch. Please revoke the request and try again.";
   }
 
-  const cliRequest = await pluginToCliRequest(provisionedRequest);
+  const cliRequest = await pluginToCliRequest(provisionedRequest, {
+    debug: args.debug,
+  });
 
   return { request: cliRequest, publicKey, privateKey };
 };
--- a/src/commands/shared.ts
+++ b/src/commands/shared.ts
@@ -12,6 +12,7 @@ import { createKeyPair } from "../common/keys";
 import { doc } from "../drivers/firestore";
 import { print2 } from "../drivers/stdio";
 import { AwsSsh } from "../plugins/aws/types";
+import { importSshKey } from "../plugins/google/ssh-key";
 import { GcpSsh } from "../plugins/google/types";
 import { SshConfig } from "../plugins/ssh/types";
 import { Authn } from "../types/identity";
@@ -59,13 +60,13 @@ export type BaseSshCommandArgs = {
   reason?: string;
   account?: string;
   provider?: "aws" | "gcloud";
+  debug?: boolean;
 };
 
 export type ScpCommandArgs = BaseSshCommandArgs & {
   source: string;
   destination: string;
   recursive?: boolean;
-  debug?: boolean;
 };
 
 export type SshCommandArgs = BaseSshCommandArgs & {
@@ -76,7 +77,6 @@ export type SshCommandArgs = BaseSshCommandArgs & {
   A?: boolean;
   arguments: string[];
   command?: string;
-  debug?: boolean;
 };
 
 const validateSshInstall = async (
@@ -187,10 +187,19 @@ export const provisionRequest = async (
     throw "Public key mismatch. Please revoke the request and try again.";
   }
 
-  return { request: provisionedRequest, publicKey, privateKey };
+  const posixAccount =
+    provisionedRequest.permission.spec.type === "gcloud"
+      ? await importSshKey(publicKey, { debug: args.debug })
+      : undefined;
+  const linuxUserName = posixAccount?.username;
+
+  return { request: provisionedRequest, publicKey, privateKey, linuxUserName };
 };
 
-export const requestToSsh = (request: Request<PluginRequest>): SshRequest => {
+export const requestToSsh = (
+  request: Request<PluginRequest>,
+  options: { gcloud: { linuxUserName?: string } }
+): SshRequest => {
   if (request.permission.spec.type === "aws") {
     const awsRequest = request as AwsSsh;
     return {
@@ -202,12 +211,16 @@ export const requestToSsh = (request: Request<PluginRequest>): SshRequest => {
       type: "aws",
     };
   } else if (request.permission.spec.type === "gcloud") {
+    if (!options.gcloud.linuxUserName) {
+      // We already throw a helpful error in ssh-key.ts importSshKey() - we should never hit this code path.
+      throw "Unexpected error: linuxUserName is required for Google Cloud destinations. Please contact support@p0.dev for assistance.";
+    }
     const gcpRequest = request as GcpSsh;
     return {
       id: gcpRequest.permission.spec.instanceName,
       projectId: gcpRequest.permission.spec.projectId,
       zone: gcpRequest.permission.spec.zone,
-      linuxUserName: gcpRequest.generated.ssh.linuxUserName,
+      linuxUserName: options.gcloud.linuxUserName,
       type: "gcloud",
     };
   } else {
--- a/src/commands/shared.ts
+++ b/src/commands/shared.ts
@@ -37,7 +37,7 @@ import yargs from "yargs";
  */
 const GRANT_TIMEOUT_MILLIS = 60e3;
 // The prefix of installed SSH accounts in P0 is the provider name
-const SUPPORTED_PROVIDERS = ["aws", "gcloud"];
+export const SUPPORTED_PROVIDERS = ["aws", "gcloud"];
 
 export type SshRequest = AwsSshRequest | GcpSshRequest;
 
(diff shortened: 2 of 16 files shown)


src/commands/shared.ts
```diff
@@ -30,19 +32,33 @@ import yargs from "yargs";
  *  to be configured
  */
 const GRANT_TIMEOUT_MILLIS = 60e3;
+// The prefix of installed SSH accounts in P0 is the provider name
+const SUPPORTED_PROVIDERS = ["aws", "gcloud"];
```
GGonryun, line 36: small nit: maybe these keys could use a type like SshRequest["type"] to prevent misspelling or using an invalid provider?

src/commands/scp.ts
```diff
@@ -49,6 +49,11 @@ export const scpCommand = (yargs: yargs.Argv) =>
           type: "string",
           describe: "The account on which the instance is located",
         })
+        .option("provider", {
```
GGonryun, line 52: small nit (we don't need to do this now) we can create a "shared options" for ssh/scp yargs, it seems like there are enough shared args to warrant some builder function for them. Maybe we can create a ticket and give it Morph to work on?

GGonryun (approved):
lgtm, just two small nits 👍

GGonryun (changes requested):
🥳

src/common/subprocess.ts
```diff
@@ -0,0 +1,66 @@
...
+**/
+import { print2 } from "../drivers/stdio";
+import { AgentArgs } from "../plugins/ssh-agent/types";
+import { spawn, SpawnOptionsWithoutStdio } from "node:child_process";
+
+/** Spawns a subprocess with given command, args, and options.
+ * May write content to its standard input.
+ * Stdout and stderr of the subprocess is printed to stderr in debug mode.
+ * The returned promise resolves with stdout or rejects with stderr of the subprocess. */
+export const asyncSpawn = async (
```
gergas3, line 19: This moved from ssh-agent.ts
The difference is now it returns stdout/stderr instead of exit code. This is helpful because when this function is called from the new gcloud path that gets the current user and the access token because:
1. we will print a nice error message if user is not logged in to gcloud (the stderr of the `gcloud ...` command)
2. we can return the access token and user as the result of awaiting this function

src/plugins/ssh/index.ts
```diff
@@ -63,6 +63,7 @@ const DEFAULT_VALIDATION_WINDOW_MS = 5e3;
  * Note that each attempt consumes ~ 1 s.
  */
 const MAX_SSH_RETRIES = 30;
+const GCP_MAX_SSH_RETRIES = 120; // GCP requires more time to propagate access
```
GGonryun, line 66: What about using a map for this?

```
const MAX_SSH_RETRIES: Record<SshRequest["type"], number> = { "aws": 30, "gcloud": 120 }

// somewhere else...
spawnSshNode({
  ...
  attemptsRemaining: MAX_SSH_RETRIES[data.type]
});
```

src/plugins/google/ssh-key.ts
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * The user account and the access token is retrieved from the gcloud CLI.
+ *
+ * Returns the posix account to use for SSH access.
+ *
+ * See https://cloud.google.com/compute/docs/oslogin/rest/v1/users/importSshPublicKey
+ */
+export const importSshKey = async (
+  publicKey: string,
+  options?: { debug?: boolean }
```
GGonryun, line 28: What if you just made debug required? `options : {debug: boolean} = {debug: false}`.

Then you could pass the options into async spawn directly: `await asyncSpawn(options, "gcloud", [...]`

src/commands/shared.ts
```diff
@@ -187,10 +187,19 @@ export const provisionRequest = async (
     throw "Public key mismatch. Please revoke the request and try again.";
   }
 
-  return { request: provisionedRequest, publicKey, privateKey };
+  const posixAccount =
+    provisionedRequest.permission.spec.type === "gcloud"
+      ? await importSshKey(publicKey, { debug: args.debug })
+      : undefined;
+  const linuxUserName = posixAccount?.username;
```
GGonryun, line 194: It feels a little unexpected to me here to return the linux username from without any mention that it's GCP specific in the output. Since this is happening last could we move it to `requestToSsh`? Is there a disadvantage to running this code in `requestToSsh` on line 214?

GGonryun (approved):
LGTM! just a few style nits.

src/commands/shared.ts
```diff
@@ -187,10 +187,19 @@ export const provisionRequest = async (
...
+
+  return { request: provisionedRequest, publicKey, privateKey, linuxUserName };
 };
 
-export const requestToSsh = (request: Request<PluginRequest>): SshRequest => {
+export const requestToSsh = (
+  request: Request<PluginRequest>,
+  options: { gcloud: { linuxUserName?: string } }
```
nbrahms, line 201: Isn't the `linuxUserName` already on the request object? Why not just dither on the request type directly?

src/common/subprocess.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    let stdout = "";
+    let stderr = "";
+
+    if (writeStdin) {
+      if (!child.stdin) return reject("Child process has no stdin");
+      child.stdin.write(writeStdin);
+    }
+
+    child.stdout.on("data", (data) => {
```
nbrahms, line 37: what's the point of this over just piping the streams?

src/common/subprocess.ts
```diff
@@ -0,0 +1,66 @@
...
+  { debug }: AgentArgs,
+  command: string,
+  args?: ReadonlyArray<string>,
+  options?: SpawnOptionsWithoutStdio,
+  writeStdin?: string
+) =>
+  new Promise<string>((resolve, reject) => {
+    const child = spawn(command, args, options);
+
+    let stdout = "";
```
nbrahms, line 29: We should at least note in the jsdoc that this will blow up with large outputs.

Ideally we'd instead write data to a stream rather than a string.

src/common/subprocess.ts
```diff
@@ -0,0 +1,66 @@
...
+    let stderr = "";
+
+    if (writeStdin) {
+      if (!child.stdin) return reject("Child process has no stdin");
+      child.stdin.write(writeStdin);
+    }
+
+    child.stdout.on("data", (data) => {
+      const str = data.toString("utf-8");
+      stdout += str;
```
nbrahms, line 39: Additionally, this takes a performance hit if there are many `on("data")` calls.

If you instead use a stream or a buffer you eliminate the constant re-allocation of memory here.

src/plugins/google/ssh-key.ts
```diff
@@ -0,0 +1,75 @@
...
+  });
+  const data: ImportSshPublicKeyResponse = await response.json();
+  if (debug) {
+    print2(
+      `Login profile for user after importing public key: ${JSON.stringify(data)}`
+    );
+  }
+  const { loginProfile } = data;
+  // Find the primary POSIX account for the user, or the first in the array
+  const posixAccount =
```
nbrahms, line 64: It seems to me that this lookup should be done on the P0 backend, rather than in the CLI. Or, it should be a request argument.

src/plugins/google/ssh-key.ts
```diff
@@ -0,0 +1,75 @@
...
+    loginProfile.posixAccounts.find((account) => account.primary) ||
+    // nosemgrep no-stringify-keys
```
nbrahms, line 66: Suggest ignoring this in the Semgrep app instead; this appears to be a false positive, rather than a deliberate violation.

src/plugins/google/ssh-key.ts
```diff
@@ -0,0 +1,75 @@
...
+    loginProfile.posixAccounts[0];
```
nbrahms, line 67: What happens if the OS mismatches the machine OS for this `PosixAccount` instance?

src/plugins/ssh/index.ts
nbrahms, line 66: How long does the SSH attempt take with GCP? I presume the jsdoc comment assumes AWS.

If the consumed time per attempt is significantly shorter, should we use a sleep instead of retrying more?

src/plugins/ssh/index.ts
```diff
@@ -63,6 +63,7 @@ const DEFAULT_VALIDATION_WINDOW_MS = 5e3;
  * Note that each attempt consumes ~ 1 s.
  */
 const MAX_SSH_RETRIES = 30;
```
nbrahms, line 65: If we differ between AWS and GCP we should indicate that this constant is for AWS.

src/plugins/ssh/index.ts
```diff
@@ -385,6 +386,9 @@ export const sshOrScp = async (
       );
     }
 
+    const maxRetries =
+      data.type === "gcloud" ? GCP_MAX_SSH_RETRIES : MAX_SSH_RETRIES;
```
nbrahms, line 390: Let's please make this an exhaustive type check. Otherwise we'll forget to update this if we add another CSP.

src/commands/ssh.ts
```diff
@@ -94,15 +94,15 @@ const sshAction = async (args: yargs.ArgumentsCamelCase<SshCommandArgs>) => {
     throw "Server did not return a request id. Please contact support@p0.dev for assistance.";
   }
 
-  const { request } = result;
+  const { request, privateKey, linuxUserName } = result;
 
   await sshOrScp(
     authn,
-    requestToSsh(request),
+    requestToSsh(request, { gcloud: { linuxUserName } }),
```
nbrahms, line 101: 🤔 it's weird that there's a mention of "gcloud" in the ssh base command. I'd expect anything GCP-related (and it's GCP, not `gcloud`, correct?) to be in the `plugins/google` directory.

src/commands/shared.ts
```diff
@@ -202,12 +211,16 @@ export const requestToSsh = (request: Request<PluginRequest>): SshRequest => {
       type: "aws",
     };
   } else if (request.permission.spec.type === "gcloud") {
```
nbrahms, line 213: This appears to violate separation of concerns.

Would make more sense I think to add SSH-specific callbacks in the two CSP plugins, and look those up from here.

src/commands/shared.ts
```diff
@@ -76,7 +77,6 @@ export type SshCommandArgs = BaseSshCommandArgs & {
   A?: boolean;
   arguments: string[];
   command?: string;
-  debug?: boolean;
 };
 
 const validateSshInstall = async (
```
nbrahms, line 82: Can we please move everything SSH-related into `ssh.ts`?

src/commands/shared.ts
```diff
@@ -187,10 +187,19 @@ export const provisionRequest = async (
     throw "Public key mismatch. Please revoke the request and try again.";
   }
 
-  return { request: provisionedRequest, publicKey, privateKey };
+  const posixAccount =
+    provisionedRequest.permission.spec.type === "gcloud"
```
nbrahms, line 191: and everything google-related into `plugins/google`

src/commands/scp.ts
```diff
@@ -84,9 +84,9 @@ const scpAction = async (args: yargs.ArgumentsCamelCase<ScpCommandArgs>) => {
   if (!result) {
     throw "Server did not return a request id. Please contact support@p0.dev for assistance.";
   }
-  const { request, privateKey } = result;
+  const { request, privateKey, linuxUserName } = result;
 
-  const data = requestToSsh(request);
+  const data = requestToSsh(request, { gcloud: { linuxUserName } });
```
nbrahms, line 89: separation of concerns here, as well

src/plugins/ssh/index.ts
gergas3: This is intentional: GCP is a special case with extra long wait time. I didn't mean to make this exhaustive.

src/plugins/ssh/index.ts
gergas3: This is a default. I will rename it to `DEFAULT_MAX_SSH_RETRIES`

src/plugins/google/ssh-key.ts
gergas3: We can't get the profile of the user in the backend without getting Directory API permissions.

src/commands/shared.ts
gergas3: For GCP requests, we are not using the request object's `linuxUserName`. Actually it's not even in the request object for GCP.

src/common/subprocess.ts
gergas3: We return the entire output once the subprocess exits. We could do it with streams internally but in the end it's still a stream that we return. Or we could return the promise of stream and read that. For the purpose of executing the gcloud commands we have (and the ssh-agent commands) this should work.

src/plugins/google/ssh-key.ts
gergas3: That's a bug! Thanks for catching.

src/plugins/ssh/index.ts
gergas3: That depends on the type of issue with the access.
iirc 
> 3. The user cannot act as the service account of the compute instance
takes the longest on the order of several seconds, while most others are really quick, under 1s.

So the nice solution would be to remember the start time and sleep based on elapsed time. I will give that a stab.

src/commands/ssh.ts
gergas3: Good point, I will move this.
`gcloud` is the constant we have been using for better or worse. True, it is the name of the CLI.
On the other hand, Google Cloud Platform was renamed to just Google Cloud some time ago, so that invalidates the GCP abbreviation? (Of course people still use it.)

src/commands/shared.ts
gergas3: Hm, I think this belongs to "shared". This "shared" file contains the pieces of logic common to ssh and scp. This function validates the P0 SSH component install which is required by both.

src/commands/shared.ts
gergas3: `importSshKey` function is in `plugins/google`.
This file is shared between scp and ssh not between clouds.
If we don't want to have branching like this I think we have to couple data with implementation, which we don't do now in general.

src/commands/scp.ts
gergas3: This object indicates that this is gcloud-specific data.
This `scp` and the `shared` module where `requestToSsh` lives are shared between ssh and scp, not between CSPs.
You had another comment in `requestToSsh` to call out to CSP-specific modules which I will do.
I'm not sure what else to do on this line.

src/commands/ssh.ts
gergas3: Actually, on second thought I think this is not that weird. I wanted to make it clear that the linuxUserName will only be used for gcloud in the called function.
I can hide this field inside PluginRequest to make it less confusing.

src/commands/scp.ts
gergas3: See https://github.com/p0-security/p0cli/pull/90#discussion_r1697685917

src/commands/ssh.ts
gergas3: I defined a new type `CliRequest` to differentiate from `PluginRequest` that we get back from the p0 backend. The `CliRequest` type has an additional field `cliLocalData` which contains the `linuxUserName` for gcp. It's undefined for aws.

src/plugins/ssh/index.ts
gergas3: In the interest of time I'm going to punt on this in the current PR.

src/commands/scp.ts
```diff
@@ -49,6 +49,11 @@ export const scpCommand = (yargs: yargs.Argv) =>
           type: "string",
           describe: "The account on which the instance is located",
         })
+        .option("provider", {
+          type: "string",
+          describe: "The cloud provider where the instance is hosted",
+          choices: ["aws", "gcloud"],
```
nbrahms, line 55:
```suggestion
          choices: SUPPORTED_PROVIDERS,
```
?

src/plugins/ssh/index.ts
```diff
@@ -0,0 +1,404 @@
...
+): Promise<number | null> {
+  return new Promise((resolve, reject) => {
+    const child = spawnChildProcess(
+      options.credential,
+      options.command,
+      options.args,
+      options.stdio
+    );
+
+    // TODO ENG-2284 support login with Google Cloud: currently return a boolean to indicate if the exception was a Google login error.
```
nbrahms, line 210: does this still apply

nbrahms (approved):

src/plugins/ssh/index.ts
gergas3: Yes, it's still a TODO to automatically login (or at least prompt a second login)

PR #91, opened 2024-07-26
Aws Ssh using aws idc

varunalla:
Addresses: ENG-2233

Allows P0 cli to authenticate to idc to allow ssh.

paired with : https://github.com/p0-security/app/pull/1852


src/commands/scp.ts
```diff
@@ -12,10 +12,10 @@ import { authenticate } from "../drivers/auth";
 import { guard } from "../drivers/firestore";
 import { sshOrScp } from "../plugins/aws/ssm";
 import {
-  ScpCommandArgs,
```
GGonryun, line 15: Was this change necessary?

src/plugins/oidc/login.ts
```diff
@@ -57,70 +71,106 @@ const authorize = async (org: OrgData, scope: string) => {
...
-  const authorizeResponse = await authorize(org, scope);
-  print2(`Please use the opened browser window to continue your P0 login.
+export const oidcLoginSteps: (
+  org: OrgData,
+  scope: string
+) => OidcLoginSteps<AuthorizeResponse, TokenResponse> = (
+  org: OrgData,
+  scope: string
+) => ({
+  setup: async () => {
```
GGonryun, line 141: nit: I would expect this to be called something like `validate` because it's not actually saving/storing any data. set up makes me think we're "installing" some data somehow.

src/plugins/oidc/login.ts
```diff
@@ -57,70 +71,106 @@ const authorize = async (org: OrgData, scope: string) => {
...
-    const response = await fetchOidcToken(org, authorize);
-    if (!response) await sleep(authorize.interval * 1e3);
-    else return response;
-  }
-  throw "Expired awaiting in-browser authorization.";
-};
+export const waitForActivation =
+  <A, T>(authorize: A & { interval: number }) =>
+  async (
+    getExpiry: (authorize: A) => number, // Aws implementation differs from standard OIDC response, need function to extract expiry
```
GGonryun, line 122: Clarity on this comment: does this mean that the AWS Identity Center is different than Okta SAML?

src/commands/scp.ts
varunalla: the import reordered.

src/plugins/oidc/login.ts
varunalla: aws oidc implementation does not conform to the standard oidc spec. especially the naming convention.

src/commands/shared.ts
```diff
@@ -165,11 +174,23 @@ export const provisionRequest = async (
 };
 
 export const requestToSsh = (request: AwsSsh): SshRequest => {
```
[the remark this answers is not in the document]
GGonryun: I wrote this because I didn't like that `generated.name` was the `role.name` and to flatten out the object so that if the `permission` changes, we only need to change the `sshRequest` in one place (here where it's getting transformed). Probably isn't all that necessary to be honest.

src/plugins/aws/ssm/index.ts
```diff
@@ -275,16 +286,32 @@ export const sshOrScp = async (
...
+      role: data.role,
+    });
+  } else {
+    throw `Unsupported login type: ${config.login?.type}`;
+  }
   return withSshAgent(cmdArgs, async () => {
     const { command, args } = createProxyCommands(data, cmdArgs, cmdArgs.debug);
 
-    if (cmdArgs.debug) {
+    // TODO: Modify commands to add the ability to get permission set commands
```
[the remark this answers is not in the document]
varunalla: the current debug instructions do not consider the oidc flow to set aws credentials and provides instructions to assume role. instructions should be added to setup the idc flow to setup sso and use those creds to ssh.

src/common/backoff.ts
```diff
@@ -0,0 +1,32 @@
...
+@p0security/cli is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
+**/
+import { sleep } from "../util";
+
+const MAX_RETRIES = 3;
+const MAX_RETRY_BACK_OFF_TIME = 10000;
+
+export async function retryWithBackOff<T>(
```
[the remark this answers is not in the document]
GGonryun: NIT: instead of hardcoding the retry with a delay we should change the function to take an object with a configurable settings for retries and delay.

```
type RetryOptions<T> = {
  operation: () => Promise<T>
  when: (error: unknown) => boolean
  max?: number
  delay?: number
}
retryWithSleep<T>(options: RetryOptions)
```

src/common/backoff.ts
```diff
@@ -0,0 +1,32 @@
...
+  cb: () => Promise<T>,
+  retryPredicate: (error: any) => boolean,
```
GGonryun, line 18: should be `error: unknown` instead of any. It's a safer alternative and provides compile-time safety.

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,199 @@
...
+        }),
+      };
+      const response = await fetch(
+        `https://oidc.${region}.amazonaws.com/client/register`,
+        init
+      );
+      return await response.json();
+    },
+    {
+      duration: 7.776e9, // the token has lifetime of 90 days
```
GGonryun, line 52: small nit: there's a handful of magic values. for example this could be a separate variable called `TOKEN_LIFETIME` and the scopes could be `AWS_IDC_SCOPES = ["sso:account:access"]`. no strong opinion here, it just makes it easier to modify configuration in the future if it's not scattered.

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,199 @@
...
+  DEVICE_GRANT_TYPE,
+  fetchOidcToken,
+  oidcLogin,
+  waitForActivation,
+} from "../../oidc/login";
+import { AwsCredentials } from "../types";
+import open from "open";
+
+export const registerClient = async (
+  region: string
```
GGonryun, line 31: should we validate the region or is it always guaranteed to be valid?

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,199 @@
...
+    );
+    print2(`Please use the opened browser window to continue your idc device authorization.
+  
+      When prompted, confirm that aws idc page displays this code:
+      
+        ${authorizeResponse.userCode}
+      
+      Waiting for authorization...
+      `);
+    void open(authorizeResponse.verificationUriComplete);
```
GGonryun, line 148: do we need `void` here? 🤔

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,199 @@
...
+  idc: { id: string; region: string }
+): Promise<AwsCredentials> =>
+  await cached(
+    `aws-idc-${args.account}-${args.permissionSet}`,
+    async () => {
+      const { region } = idc;
+      // fetch aws client secrets
+      const clientSecrets = await registerClient(region);
+      const oidcResponse = await cached(
+        `aws-idc-device-authorization`,
```
GGonryun, line 174: nit: we should use single quotes for strings that don't have templates.

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,199 @@
...
+      // fetch device authorization
+      const credentials = await exchangeForAWSCredentials(oidcResponse, idc, {
+        account: args.account,
+        permissionSet: args.permissionSet,
+      });
+      return {
+        AWS_ACCESS_KEY_ID: credentials.roleCredentials.accessKeyId,
+        AWS_SECRET_ACCESS_KEY: credentials.roleCredentials.secretAccessKey,
+        AWS_SESSION_TOKEN: credentials.roleCredentials.sessionToken,
+        AWS_SECURITY_TOKEN: credentials.roleCredentials.securityToken ?? "", // portal does not return this value
```
GGonryun, line 193: I think we set the security token because it's a legacy value that can override `AWS_SESSION_TOKEN`.

It should be set to `AWS_SECURITY_TOKEN: credentials.roleCredentials.sessionToken`. 

See Okta implementation here: https://github.com/p0-security/p0cli/blob/5a4a9d7c731b5590e0756c7e66fc89d619bd6160/src/plugins/aws/assumeRole.ts#L36-L37

We ran into this problem with Venkat before, @gergas3 has more context on this.

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,199 @@
...
+        expiresAt: credentials.roleCredentials.expiration,
+      };
+    },
+    { duration: 3600e3 },
```
GGonryun, line 197: DRY: maybe we should extract `3600e3` as TOKEN_LIFETIME somewhere since it appears here and on line 180?

src/plugins/oidc/login.ts
```diff
@@ -57,70 +71,106 @@ const authorize = async (org: OrgData, scope: string) => {
...
+const oidcRequestBuilder = (org: OrgData, authorize: AuthorizeResponse) => {
+  if (org.providerType === undefined) {
+    throw "Login requires a configured provider type.";
+  }
+  validateProviderDomain(org);
+
+  return {
+    url:
+      org.providerType === "okta"
+        ? `https:${org.providerDomain}/oauth2/v1/token`
```
GGonryun, line 99: `https:${org.providerDomain}/oauth2/v1/token`

This template looks wrong, shouldn't it be `https://`? 🤔  like below on line 101?

src/plugins/aws/idc/index.ts
varunalla: setting it to session token.

src/plugins/aws/idc/index.ts
varunalla: we do not need evaluate the result of the promise to open the browser window.

src/plugins/aws/ssm/index.ts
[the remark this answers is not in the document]
nbrahms: Hmm, I'm still a bit confused. Maybe we can discuss in person?

src/plugins/aws/idc/index.ts
nbrahms: Seems like it would be useful to make a small library with constants for `SECONDS`, `MINUTES`, `HOURS`, etc., and then write:
```suggestion
      duration: 90 * DAYS,
```
here

src/plugins/aws/idc/index.ts
nbrahms: @GGonryun we use `void` to indicate explicitly that we are not handling the promise; errors in the promise will crash the program.

src/plugins/aws/idc/index.ts
nbrahms: We can also use `1 * HOURS`

src/common/retry.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @param when condition to retry
+ * @param retries number of retries
+ * @param delay time to wait before retrying
+ * @returns
+ */
+export async function retryWithSleep<T>(
+  operation: () => Promise<T>,
+  when: (error: unknown) => boolean,
+  retries: number = MAX_RETRIES,
+  delay: number = MAX_RETRY_BACK_OFF_TIME
```
GGonryun, line 28: small nit on types
```suggestion
  retries = MAX_RETRIES,
  delay = MAX_RETRY_BACK_OFF_TIME
```

src/commands/shared.ts
```diff
@@ -30,34 +36,51 @@ import yargs from "yargs";
...
   region: string;
   id: string;
+  type: "aws";
 };
 
-export type AwsSshRoleRequest = BaseSshRequest & { role: string; type: "role" };
-export type AwsSshIdcRequest = BaseSshRequest & {
+export type AwsSshRoleRequest = BaseAwsSshRequest & {
+  role: string;
+  subType: "role";
```
gergas3, line 54: @varunalla This was called `type` before this merge commit. I renamed it to `subType` because type have values "aws", "gcloud".

I don't see this subType used anywhere except when we build the debug commands. When we log in to AWS we check the actual install config.

nbrahms: @gergas3 in the backend this will become `access`, if we want to adopt that here. I think it's a bit more descriptive than merely `subType`.

src/commands/aws/role.ts
```diff
@@ -46,7 +46,7 @@ export const role = (yargs: yargs.Argv<{ account: string | undefined }>) =>
       .demandCommand(1)
   );
 
-const isFederatedLogin = (
+export const isFederatedLogin = (
```
nbrahms, line 49: This doesn't appear to be used anywhere else; we should leave it unexported.

src/common/retry.ts
nbrahms, line 28:
```suggestion
  delayMs: number = MAX_RETRY_BACK_OFF_TIME
```

is more descriptive

src/common/retry.ts
```diff
@@ -0,0 +1,41 @@
...
+**/
+import { sleep } from "../util";
+
+const MAX_RETRIES = 3;
+const MAX_RETRY_BACK_OFF_TIME = 10000;
+
+/**
+ * Retries an operation with a delay between retries
+ * @param operation operation to retry
+ * @param when condition to retry
```
nbrahms, line 19: Suggestion:
```suggestion
 * @param shouldRetry predicate to evaluate on error; will retry only if this is true
```

`shouldRetry` is also more descriptive of a name

src/common/retry.ts
```diff
@@ -0,0 +1,41 @@
...
+  retries: number = MAX_RETRIES,
+  delay: number = MAX_RETRY_BACK_OFF_TIME
+): Promise<T> {
+  try {
+    return await operation();
+  } catch (error: any) {
+    if (when(error)) {
+      if (retries > 0) {
+        await sleep(delay);
+        return retryWithSleep(operation, when, retries - 1);
```
nbrahms, line 36:
```suggestion
        return await retryWithSleep(operation, when, retries - 1);
```

js/ts is weird in that this does exactly the same thing, but it's more descriptive to await here.

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,194 @@
...
+import {
+  AWSAuthorizeResponse,
+  AWSClientInformation,
+  AWSTokenResponse,
+} from "../../../types/aws/oidc";
+import { OidcLoginStepHelpers } from "../../../types/oidc";
+import { DEVICE_GRANT_TYPE, oidcLogin } from "../../oidc/login";
+import { AwsCredentials } from "../types";
+
+const AWS_TOKEN_EXPIRY = 3600e3;
```
nbrahms, line 23: suggestion: add time constants, and use
```suggestion
const AWS_TOKEN_EXPIRY = 1 * HOURS;
```
here

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,194 @@
...
+const AWS_CLIENT_TOKEN_EXPIRY = 7.776e9;
```
nbrahms, line 24: Similarly, use `90 * DAYS` here (which is actually slightly longer than the constant you have)

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,194 @@
...
+          scopes: AWS_SSO_SCOPES,
+        }),
+      };
+      const response = await fetch(
+        `https://oidc.${region}.amazonaws.com/client/register`,
+        init
+      );
+      return await response.json();
+    },
+    { duration: AWS_CLIENT_TOKEN_EXPIRY }, // the token has lifetime of 90 days
```
nbrahms, line 48: This comment should go next to the constant definition

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,194 @@
...
+        method: "POST",
+        body: JSON.stringify({
+          clientId,
+          clientSecret,
+          startUrl: `https://${id}.awsapps.com/start`,
+        }),
+      },
+      url: `https://oidc.${region}.amazonaws.com/device_authorization`,
+    };
+  };
```
nbrahms, line 71: Nit
```suggestion
  const oidcAuthorizationRequest = () => ({
      init: {
        method: "POST",
        body: JSON.stringify({
          clientId,
          clientSecret,
          startUrl: `https://${id}.awsapps.com/start`,
        }),
      },
      url: `https://oidc.${region}.amazonaws.com/device_authorization`,
    });
```

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,194 @@
...
+  const buildIdcTokenRequest = (authorizeResponse: AWSAuthorizeResponse) => {
+    return {
```
nbrahms, line 73: Nit:
```suggestion
  const buildIdcTokenRequest = (authorizeResponse: AWSAuthorizeResponse) => ({
```

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,194 @@
...
+   * @param oidcResponse oidc token response fot he oidc /token endpoint
+   * @param request accountId and permissionSet to exchange for AWS credentials
+   * @returns
+   */
+  const exchangeForAwsCredentials = async (
+    oidcResponse: AWSTokenResponse,
+    request: { accountId?: string; permissionSet: string }
+  ) => {
+    // There is a delay in between aws issuing sso token and it to be exchanged for AWS credentials
+    // encountered unauthorized error when exchanging token immediately
```
nbrahms, line 97: Grammar nits:
```suggestion
    // There is a delay in between aws issuing the sso token and it being available for exchange for AWS credentials
    // When exchanging token immediately, an "unauthorized" may will be thrown, so retry with sleep.
```

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,194 @@
...
+      async () => {
+        const init = {
+          method: "GET",
+          headers: {
+            "x-amz-sso_bearer_token": oidcResponse.accessToken,
+          },
+        };
+        const { accountId, permissionSet } = request;
+        if (accountId === undefined)
+          throw new Error("Could not find account Id for AWS SSO");
```
nbrahms, line 108: Suggestion:
```suggestion
          throw new Error("Could not find an AWS account ID for this access request");
```

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,194 @@
...
+        const params = new URLSearchParams();
+        params.append("account_id", accountId);
+        params.append("role_name", permissionSet);
+        const response = await fetch(
+          `https://portal.sso.${region}.amazonaws.com/federation/credentials?${params.toString()}`,
+          init
+        );
+        if (!response.ok)
+          throw new Error(
+            `Failed to fetch AWS credentials: ${response.statusText}`
```
nbrahms, line 118: `statusText` is probably incomplete (it'll just say something like "BAD_REQUEST"). Instead, you probably want
```suggestion
            `Failed to fetch AWS credentials: ${response.statusText}: ${await response.text()}`
```

src/plugins/aws/idc/index.ts
```diff
@@ -0,0 +1,194 @@
...
+    exchangeForAwsCredentials,
+  };
+};
+
+/**
+ * Returns AWS credentials for the specified account and permission set for the authorized user
+ * @param args accountId, permissionSet and idc to assume role associated with the permission set
+ * @returns
+ */
+export const assumeRoleWithIdc = async (args: {
```
nbrahms, line 151: nice, looks good

src/plugins/aws/ssh.ts
```diff
@@ -12,12 +12,23 @@ import { SshRequest } from "../../commands/shared";
...
-    accountId: request.permission.spec.accountId,
-    region: request.permission.spec.region,
-    role: request.generated.name,
-    linuxUserName: request.generated.ssh.linuxUserName,
-    type: "aws",
-  };
+  const { permission, generated } = request;
+  const {
+    spec: { instanceId, accountId, region },
+  } = permission;
```
nbrahms, line 18:
```suggestion
  const { instanceId, accountId, region } = permission.spec;
```

src/plugins/aws/ssh.ts
```diff
@@ -12,12 +12,23 @@ import { SshRequest } from "../../commands/shared";
...
+  const {
+    idc,
+    ssh: { linuxUserName },
+    name,
+  } = generated;
```
nbrahms, line 23:
```suggestion
  const { idc, ssh, name } = generated;
  const { linuxUserName } = ssh;
```

src/plugins/oidc/login.ts
```diff
@@ -11,44 +11,31 @@ You should have received a copy of the GNU General Public License along with @p0
...
-      ? `https:${org.providerDomain}/oauth2/v1/device/authorize`
-      : org.providerType === "ping"
-        ? `https://${org.providerDomain}/${org.environmentId}/as/device_authorization`
-        : throwAssertNever(org.providerType);
+export const authorize = async <T>(
+  request: {
+    url: string;
+    init: RequestInit;
+  },
+  validateResponse?: (response: Response) => Promise<Response>
```
nbrahms, line 33: In which situation is this ever undefined? It might be better to force the consumer to explicitly pass an async noop instead of ommitting this, to prevent unintended bugs.

src/plugins/oidc/login.ts
```diff
@@ -57,70 +44,142 @@ const authorize = async (org: OrgData, scope: string) => {
...
       const data = await response.json();
       if (data.error === "authorization_pending") return undefined;
     }
     await validateResponse(response);
   }
-  return (await response.json()) as TokenResponse;
+  return (await response.json()) as T;
+};
+
+const providerType: (org: OrgData) => NonNullable<string> = (org) => {
```
nbrahms, line 63:
```suggestion
const providerType: (org: OrgData) => string = (org) => {
```

`string` is already non-nullable

src/plugins/oidc/login.ts
```diff
@@ -57,70 +44,142 @@ const authorize = async (org: OrgData, scope: string) => {
...
  * Returns the OIDC token after completion.
  */
-const waitForActivation = async (
-  org: OrgData,
-  authorize: AuthorizeResponse
+export const waitForActivation = async <A, T>(
+  authorize: A,
+  extractExpiryInterval: (authorize: A) => {
+    expires_in: number;
+    interval: number;
```
nbrahms, line 78: Please add a comment to indicate that this is in seconds, not milliseconds

src/plugins/oidc/login.ts
```diff
@@ -57,70 +44,142 @@ const authorize = async (org: OrgData, scope: string) => {
...
+export const oidcLoginSteps = (
+  org: OrgData,
+  scope: string,
+  urls: () => { deviceAuthorizationUrl: string; tokenUrl: string }
+) => {
+  const { deviceAuthorizationUrl, tokenUrl } = urls();
+
+  const buildOidcAuthorizeRequest = () => {
+    if (org.providerType === undefined) {
+      throw "Login requires a configured provider type.";
```
nbrahms, line 101: What is the user supposed to do with this? Also, do they know what a "provider type" is?

We should direct the user to contact their P0 admin:

```suggestion
      throw "Your organization's login configuration does not support this access. Your P0 admin will need to install a supported OIDC provider in order for you to use this command.";
```

src/plugins/oidc/login.ts
```diff
@@ -57,70 +44,142 @@ const authorize = async (org: OrgData, scope: string) => {
...
+    }
+
+    validateProviderDomain(org);
+    // This is the "org" authorization server; the okta.apps.* scopes are not
+    // available with custom authorization servers
```
nbrahms, line 106: This is generic OIDC code. Why is there anything about Okta here? Shouldn't that live in `okta/login`?

src/plugins/oidc/login.ts
```diff
@@ -57,70 +44,142 @@ const authorize = async (org: OrgData, scope: string) => {
...
+        body: urlEncode({
+          client_id: org.clientId,
+          scope,
+        }),
+      },
+      url: deviceAuthorizationUrl,
+    };
+  };
+  const buildOidcTokenRequest = (authorize: AuthorizeResponse) => {
+    if (org.providerType === undefined) {
```
nbrahms, line 120: Please share this code with that on line 101. We should maintain error messages consistently.

src/plugins/oidc/login.ts
```diff
@@ -57,70 +44,142 @@ const authorize = async (org: OrgData, scope: string) => {
...
+  } as OidcLoginStepHelpers<AuthorizeResponse>;
+};
+
 /** Logs in to an Identity Provider via OIDC */
-export const oidcLogin = async (org: OrgData, scope: string) => {
-  if (org.providerType === undefined) {
-    throw "Login requires a configured provider type.";
-  }
-  const authorizeResponse = await authorize(org, scope);
+export const oidcLogin = async <A, T>(context: OidcLoginStepHelpers<A>) => {
```
nbrahms, line 155: naming nit:
```suggestion
export const oidcLogin = async <A, T>(steps: OidcLoginStepHelpers<A>) => {
```

src/plugins/oidc/login.ts
```diff
@@ -57,70 +44,142 @@ const authorize = async (org: OrgData, scope: string) => {
...
-When prompted, confirm that ${capitalize(org.providerType)} displays this code:
-
-  ${authorizeResponse.user_code}
 
-Waiting for authorization...
-`);
-  void open(authorizeResponse.verification_uri_complete);
-  const oidcResponse = await waitForActivation(org, authorizeResponse);
-  return oidcResponse;
+    When prompted, confirm that ${capitalize(providerType)} displays this code:
```
nbrahms, line 173: We probably want a mapping of supported provider types to human-friendly labels. E.g., I think the P0-internal `providerType` for "Entra ID" is `azure-ad`; merely capitalizing this will likely confuse the end user.

src/plugins/okta/login.ts
```diff
@@ -76,7 +80,17 @@ const fetchSamlResponse = async (
 
 /** Logs in to Okta via OIDC */
 export const oktaLogin = async (org: OrgData) =>
-  oidcLogin(org, "openid email profile okta.apps.sso");
+  oidcLogin<AuthorizeResponse, TokenResponse>(
+    oidcLoginSteps(org, "openid email profile okta.apps.sso", () => {
+      if (org.providerType !== "okta") {
+        throw `Invalid provider type ${org.providerType}`;
```
nbrahms, line 86: What is the user supposed to do with this information? Or does hitting this imply a code bug?

If it's the latter, we should emit a stack trace:
```suggestion
        throw new Error(`Invalid provider type ${org.providerType} (expected "okta")`);
```

src/plugins/ping/login.ts
```diff
@@ -8,9 +8,20 @@ This file is part of @p0security/cli
...
-import { oidcLogin } from "../oidc/login";
+import { oidcLogin, oidcLoginSteps } from "../oidc/login";
 
 /** Logs in to PingOne via OIDC */
 export const pingLogin = async (org: OrgData) =>
-  oidcLogin(org, "openid email profile");
+  oidcLogin<AuthorizeResponse, TokenResponse>(
+    oidcLoginSteps(org, "openid email profile", () => {
+      if (org.providerType !== "ping" || org.providerType === undefined) {
+        throw `Invalid provider type ${org.providerType}`;
```
nbrahms, line 20: Ditto here

src/types/oidc.ts
```diff
@@ -50,3 +50,21 @@ export type TokenErrorResponse = {
     | "not found"
     | "slow_down";
 };
+
+export type OidcLoginStepHelpers<A> = {
```
nbrahms, line 54: suggestion:
```suggestion
export type OidcLoginSteps<A> = {
```

nbrahms (approved):
Logic looks reasonable.

However, we should probably spend some time improving the user experience when edge cases / errors are encountered.

src/plugins/oidc/login.ts
varunalla: makes sense

src/plugins/aws/idc/index.ts
varunalla: it should be valid, given this is the response from the idc call.

PR #92, opened 2024-08-01
ssh: Organize code

nbrahms:
Organizes:
- SSH-specific shared code into its own file
- AWS- and GCP-specific code into their own plugin files (as FooSshProvider objects)

There's still entanglement of AWS-specific code within the base SSH plugin code, but I've left organizing that as separate work.

This is a pure code move accomplished using vscode.


gergas3 (approved):
lgtm! 🙏

GGonryun (approved):

PR #95, opened 2024-08-07
Pre-test for sudo in gcloud ssh

gergas3:
Previously, we were not able to verify if sudo access has propagated to Google Cloud VMs, specifically in the case when the user already had regular access and then requested sudo access. This is because the IAM permissions for regular access already exist, so we successfully logged in the user to the VM. However, the new IAM permissions to allow sudo have not propagated yet, so the user was not able to successfully execute any sudo commands until later.

This PR addresses the issue by executing a pre-test ssh command that runs `sudo -v` in non-interactive mode to verify that the user _can_ sudo. Once the test completes we go on to log the user in.

Also refactors code to move aws and gcloud checks from `plugins/ssh/index.ts`.

Completes ENG-2292

### DIFF
--- a/src/plugins/aws/ssh.ts
+++ b/src/plugins/aws/ssh.ts
@@ -86,8 +86,8 @@ export const awsSshProvider: SshProvider<
     ];
   },
   reproCommands: (request) => {
-    // TODO: Modify commands to add the ability to get permission set commands
-    if (request.type === "aws" && request.access !== "idc") {
+    // TODO: Add manual commands for IDC login
+    if (request.access !== "idc") {
       return [
         `eval $(p0 aws role assume ${request.role} --account ${request.accountId})`,
       ];
--- a/src/plugins/aws/ssh.ts
+++ b/src/plugins/aws/ssh.ts
@@ -9,12 +9,33 @@ This file is part of @p0security/cli
 You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
 **/
 import { SshProvider } from "../../types/ssh";
-import { AwsSshPermissionSpec, AwsSshRequest } from "./types";
+import { throwAssertNever } from "../../util";
+import { assumeRoleWithOktaSaml } from "../okta/aws";
+import { getAwsConfig } from "./config";
+import { assumeRoleWithIdc } from "./idc";
+import { ensureSsmInstall } from "./ssm/install";
+import {
+  AwsCredentials,
+  AwsSshIdcRequest,
+  AwsSshPermissionSpec,
+  AwsSshRequest,
+  AwsSshRoleRequest,
+} from "./types";
+
+/** Maximum number of attempts to start an SSH session
+ *
+ * Each attempt consumes ~ 1 s.
+ */
+const MAX_SSH_RETRIES = 30;
+
+/** The name of the SessionManager port forwarding document. This document is managed by AWS.  */
+const START_SSH_SESSION_DOCUMENT_NAME = "AWS-StartSSHSession";
 
 export const awsSshProvider: SshProvider<
   AwsSshPermissionSpec,
   undefined,
-  AwsSshRequest
+  AwsSshRequest,
+  AwsCredentials
 > = {
   requestToSsh: (request) => {
     const { permission, generated } = request;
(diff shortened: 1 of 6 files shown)


src/plugins/google/ssh.ts
```diff
@@ -35,4 +41,36 @@ export const gcpSshProvider: SshProvider<
...
+      // but mention on page https://cloud.google.com/iap/docs/tcp-by-host
+      // and also found in `gcloud ssh --dry-run` output
+      "--listen-on-stdin",
+      `--zone=${request.zone}`,
+      `--project=${request.projectId}`,
+    ];
+  },
+  reproCommands: () => undefined, // TODO @ENG-2284 support login with Google Cloud
+  preTestAccessPropagationArgs: (cmdArgs) => {
+    if (cmdArgs.sudo || ("command" in cmdArgs && cmdArgs.command === "sudo")) {
```
GGonryun, line 63: nit: is this sudo check happening anywhere else if so, we could use a helper function like: `isSudoCommand(cmdArgs)`

GGonryun (approved):
LGTM! just two tiny nits. I really like how the new ssh shared interface is coming along.

src/plugins/aws/ssh.ts
```diff
@@ -33,4 +54,47 @@ export const awsSshProvider: SshProvider<
...
+      "--target",
+      "%h",
+      "--document-name",
+      START_SSH_SESSION_DOCUMENT_NAME,
+      "--parameters",
+      '"portNumber=%p"',
+    ];
+  },
+  reproCommands: (request) => {
+    // TODO: Modify commands to add the ability to get permission set commands
```
nbrahms, line 89: I'm confused by this TODO. What's the thing we need to do?

src/plugins/aws/ssh.ts
```diff
@@ -33,4 +54,47 @@ export const awsSshProvider: SshProvider<
...
+    if (request.type === "aws" && request.access !== "idc") {
```
nbrahms, line 90: Seems like `request.type` should be guaranteed to equal `aws` already, right?

src/plugins/ssh/index.ts
```diff
@@ -55,22 +45,13 @@ const DESTINATION_READ_ERROR =
   /Error while connecting \[4010: 'destination read failed'\]/;
 const GOOGLE_LOGIN_MESSAGE =
   /You do not currently have an active account selected/;
+const SUDO_MESSAGE = /Sorry, user .+ may not run sudo on .+/; // The output of `sudo -v` when the user is not allowed to run sudo
```
nbrahms, line 48: Is this universal across unix-style operating systems? Or is this just validated on a single version of a single OS?

src/types/ssh.ts
```diff
@@ -38,12 +40,28 @@ export type SshProvider<
...
+  C extends object | undefined = undefined, // credentials object
 > = {
   requestToSsh: (request: CliPermissionSpec<PR, O>) => SR;
   toCliRequest: (
     request: Request<PR>,
     options?: { debug?: boolean }
   ) => Promise<Request<CliSshRequest>>;
+  cloudProviderLogin: (authn: Authn, request: SR) => Promise<C>;
+  proxyCommand: (request: SR) => string[];
+  reproCommands: (request: SR) => string[] | undefined;
```
nbrahms, line 52: Can you please add jsdoc to each of these? I'm confused as to what `reproCommands` etc. are meant to be.

src/plugins/ssh/index.ts
```diff
@@ -245,53 +224,19 @@ async function spawnSshNode(
...
-const createProxyCommands = (
+const createCommand = (
   data: SshRequest,
-  args: ScpCommandArgs | SshCommandArgs,
-  debug?: boolean
+  args: CommandArgs,
+  proxyCommand: string[]
 ) => {
-  let proxyCommand;
-  if (data.type === "aws") {
```
nbrahms, line 260: yay

src/plugins/ssh/index.ts
```diff
@@ -345,60 +290,79 @@ const transformForShell = (args: string[]) => {
...
-  const { config } = await getAwsConfig(authn, data.accountId);
-  if (!config.login?.type || config.login?.type === "iam") {
-    throw "This account is not configured for SSH access via the P0 CLI";
+/** Construct another command to use for testing access propagation prior to actually logging in the user to the ssh session */
+const preTestAccessPropagationIfNeeded = async (
+  request: SshRequest,
+  cmdArgs: CommandArgs,
+  proxyCommand: string[],
+  credential: AwsCredentials | undefined
+) => {
```
nbrahms, line 299: We can improve the typing here, and fully remove the dependency on AWS in the generic SSH code:

```suggestion
const preTestAccessPropagationIfNeeded = async <P extends SshProvider<any, any, any, any>>(
  request: SshRequest,
  cmdArgs: CommandArgs,
  proxyCommand: string[],
  credential: P extends SshProvider<infer _PR, infer _O, infer _SR, infer C> ? C : undefined
) => {
```

src/plugins/ssh/index.ts
```diff
@@ -345,60 +290,79 @@ const transformForShell = (args: string[]) => {
...
+  const sshProvider = SSH_PROVIDERS[request.type];
```
nbrahms, line 300: I'd probably pass the provider itself to further enhance the type safety.

nbrahms (approved):
👍 Nice improvement

src/plugins/aws/ssh.ts
gergas3: changed!

src/plugins/ssh/index.ts
gergas3: I only validated on one operating system. Should behave the same way? It's part of the sudo command itself.

src/plugins/ssh/index.ts
gergas3: Nice!

PR #96, opened 2024-08-16
Add command for automated access to EKS clusters

p0-andrewa:
This PR implements support for fully automated access to an AWS EKS cluster for AWS accounts configured to use IAM Identity Center or federated access via Okta SAML. The command, `p0 kubeconifig` will:

- Confirm that required tooling, specifically the AWS CLI and `kubectl`, are installed, and provide installation assistance if not
- Make a P0 request to the given cluster and wait for access to be granted
- If using IDC, perform an SSO login to IDC
- For both IDC and federated, write the temporary AWS credentials to a profile in the `~/.aws/credentials` file, using a unique but stable profile for each EKS cluster to allow for re-running the command to refresh credentials
    - Makes use of an INI parser and atomic file writes to guard against corruption issues and prevent stomping on other credentials
- Call `aws eks update-kubeconfig` to automatically generate or update the `kubectl` configuration
- Call `kubectl config use-context` to activate the configuration

Once done, the user can simply run `kubectl` commands to interact seamlessly with the cluster.

Tested locally against my testing EKS cluster, in both AWS IDC and AWS Federated (Okta SAML) modes. Also tested behavior when `kubectl` command is missing.

Standard flow from no access to able to access:
Standard flow

Example flow when `kubectl` not detected. Note that the `kubectl` download URL correctly reflects the system processor architecture:
Missing kubectl

### DIFF
--- a/src/commands/aws/files.ts
+++ b/src/commands/aws/files.ts
@@ -9,9 +9,9 @@ This file is part of @p0security/cli
 You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
 **/
 import { AwsCredentials } from "../../plugins/aws/types";
-import { EncodeOptions, parse, stringify } from "ini";
-import fs from "node:fs/promises";
-import os from "node:os";
+import * as ini from "ini";
+import * as fs from "node:fs/promises";
+import * as os from "node:os";
 import path from "node:path";
 import tmp from "tmp-promise";
 
@@ -29,11 +29,12 @@ export const AWS_CREDENTIALS_FILE = path.join(AWS_CONFIG_PATH, "credentials");
  * @returns Arbitrary object representing the contents of the file, or an empty
  * object if the file is empty or does not exist
  */
-export const readIniFile = async (path: string): Promise<any> => {
-  let data;
-
+export const readIniFile = async (
+  path: string
+): Promise<{ [key: string]: any }> => {
   try {
-    data = await fs.readFile(path, { encoding: "utf-8" });
+    const data = await fs.readFile(path, { encoding: "utf-8" });
+    return data ? ini.parse(data) : {};
   } catch (err: any) {
     if (err.code === "ENOENT") {
       return {};
@@ -41,12 +42,6 @@ export const readIniFile = async (path: string): Promise<any> => {
 
     throw err;
   }
-
-  if (!data) {
-    return {};
-  }
-
-  return parse(data);
 };
 
 /**
@@ -65,9 +60,9 @@ export const readIniFile = async (path: string): Promise<any> => {
 export const atomicWriteIniFile = async (
   path: string,
   obj: any,
-  iniEncodeOptions?: EncodeOptions
+  iniEncodeOptions?: ini.EncodeOptions
 ): Promise<void> => {
-  const data = stringify(obj, iniEncodeOptions);
+  const data = ini.stringify(obj, iniEncodeOptions);
 
   // Permissions will be moved along with the file
   const { path: tmpPath } = await tmp.file({ mode: 0o600, prefix: "p0cli-" });
--- a/src/commands/aws/files.ts
+++ b/src/commands/aws/files.ts
@@ -1,0 +1,106 @@
+/** Copyright © 2024-present P0 Security
+
+This file is part of @p0security/cli
+
+@p0security/cli is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 3 of the License.
+
+@p0security/cli is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
+**/
+import { AwsCredentials } from "../../plugins/aws/types";
+import { EncodeOptions, parse, stringify } from "ini";
+import fs from "node:fs/promises";
+import os from "node:os";
+import path from "node:path";
+import tmp from "tmp-promise";
+
+const AWS_CONFIG_PATH = path.join(os.homedir(), ".aws");
+export const AWS_CONFIG_FILE = path.join(AWS_CONFIG_PATH, "config");
+export const AWS_CREDENTIALS_FILE = path.join(AWS_CONFIG_PATH, "credentials");
+
+// Reference documentation: https://docs.aws.amazon.com/sdkref/latest/guide/file-format.html
+
+/**
+ * Reads in an AWS CLI configuration file, which is formatted as INI text, and
+ * returns an arbitrary object representing the contents.
+ *
+ * @param path Path of the file to read
+ * @returns Arbitrary object representing the contents of the file, or an empty
+ * object if the file is empty or does not exist
+ */
+export const readIniFile = async (path: string): Promise<any> => {
+  let data;
+
+  try {
+    data = await fs.readFile(path, { encoding: "utf-8" });
+  } catch (err: any) {
+    if (err.code === "ENOENT") {
+      return {};
+    }
+
+    throw err;
+  }
+
+  if (!data) {
+    return {};
+  }
+
+  return parse(data);
+};
+
+/**
+ * This function writes an arbitrary object as INI-formatted text to a file
+ * atomically by first writing the data to a temporary file then moving the
+ * temporary file on top of the target file. This minimizes the chance that an
+ * exception, system crash, or other similar event will leave the file in a
+ * corrupted state; this is important since we're mucking around with the AWS
+ * CLI's configuration files.
+ *
+ * @param path Path of the (permanent) file to write to
+ * @param obj Arbitrary object to convert to INI-formatted text and write to the
+ * file
+ * @param iniEncodeOptions Options to pass to the INI encoding library
+ */
+export const atomicWriteIniFile = async (
+  path: string,
+  obj: any,
+  iniEncodeOptions?: EncodeOptions
+): Promise<void> => {
+  const data = stringify(obj, iniEncodeOptions);
+
+  // Permissions will be moved along with the file
+  const { path: tmpPath } = await tmp.file({ mode: 0o600, prefix: "p0cli-" });
+
+  await fs.writeFile(tmpPath, data, { encoding: "utf-8" });
+  await fs.rename(tmpPath, path);
+};
+
+export const writeAwsTempCredentials = async (
+  profileName: string,
+  awsCredentials: AwsCredentials
+) => {
+  const credentials = await readIniFile(AWS_CREDENTIALS_FILE);
+
+  credentials[profileName] = {
+    aws_access_key_id: awsCredentials.AWS_ACCESS_KEY_ID,
+    aws_secret_access_key: awsCredentials.AWS_SECRET_ACCESS_KEY,
+    aws_session_token: awsCredentials.AWS_SESSION_TOKEN,
+  };
+
+  // The credentials file is formatted with whitespace before and after the `=`
+  await atomicWriteIniFile(AWS_CREDENTIALS_FILE, credentials, {
+    whitespace: true,
+  });
+};
+
+export const writeAwsConfigProfile = async (
+  profileName: string,
+  profileConfig: any
+) => {
+  const config = await readIniFile(AWS_CONFIG_FILE);
+
+  config[`profile ${profileName}`] = profileConfig;
+
+  await atomicWriteIniFile(AWS_CONFIG_FILE, config);
+};
--- a/src/commands/kubeconfig.ts
+++ b/src/commands/kubeconfig.ts
@@ -11,7 +11,7 @@ You should have received a copy of the GNU General Public License along with @p0
 import { retryWithSleep } from "../common/retry";
 import { authenticate } from "../drivers/auth";
 import { guard } from "../drivers/firestore";
-import { print2 } from "../drivers/stdio";
+import { Ansi, print2 } from "../drivers/stdio";
 import {
   awsCloudAuth,
   profileName,
@@ -150,6 +150,13 @@ const kubeconfigAction = async (
   print2(
     "Access granted and kubectl configured successfully. Re-run this command to refresh access if credentials expire."
   );
+
+  if (process.env.AWS_ACCESS_KEY_ID) {
+    print2(
+      `${Ansi.Yellow}Warning: AWS credentials were detected in your environment, which may cause kubectl errors. ` +
+        `To avoid issues, unset with \`unset AWS_ACCESS_KEY_ID\`.${Ansi.Reset}`
+    );
+  }
 };
 
 /**
--- a/src/commands/kubeconfig.ts
+++ b/src/commands/kubeconfig.ts
@@ -1,0 +1,248 @@
+/** Copyright © 2024-present P0 Security
+
+This file is part of @p0security/cli
+
+@p0security/cli is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 3 of the License.
+
+@p0security/cli is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
+**/
+import { retryWithSleep } from "../common/retry";
+import { authenticate } from "../drivers/auth";
+import { guard } from "../drivers/firestore";
+import { print2 } from "../drivers/stdio";
+import { getAwsConfig } from "../plugins/aws/config";
+import {
+  awsCloudAuth,
+  profileName,
+  requestAccessToCluster,
+  getAndValidateK8sIntegration,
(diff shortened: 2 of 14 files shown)


src/commands/aws/files.ts
```diff
@@ -0,0 +1,106 @@
...
+export const writeAwsTempCredentials = async (
+  profileName: string,
+  awsCredentials: AwsCredentials
+) => {
+  const credentials = await readIniFile(AWS_CREDENTIALS_FILE);
+
+  credentials[profileName] = {
+    aws_access_key_id: awsCredentials.AWS_ACCESS_KEY_ID,
+    aws_secret_access_key: awsCredentials.AWS_SECRET_ACCESS_KEY,
+    aws_session_token: awsCredentials.AWS_SESSION_TOKEN,
```
GGonryun, line 88: we also have to write the: `AWS_SECURITY_TOKEN`, it's a legacy value that gets used by other applications. it can be set to `AWS_SESSION_TOKEN` value.

https://github.com/p0-security/p0cli/blob/2ae3841b9ae920cbee4f0dd895964270af85ab31/src/plugins/aws/assumeRole.ts#L37

src/plugins/kubeconfig/install.ts
```diff
@@ -0,0 +1,70 @@
...
+        // or whether it's an officially-supported architecture.
+        return kubectlInstallCommandsDarwin();
+      },
+    },
+  },
+};
+
+export const ensureEksInstall = async () => {
+  return await ensureInstall(EksItems, EksInstall);
+};
```
GGonryun, line 70:
```suggestion
export const ensureEksInstall = async () => await ensureInstall(EksItems, EksInstall)
```

src/plugins/kubeconfig/index.ts
```diff
@@ -0,0 +1,138 @@
...
+    );
+  }
+
+  return await waitForProvisioning<K8sPermissionSpec>(authn, id);
+};
+
+export const profileName = (eksCluterName: string): string =>
+  `p0cli-managed-eks-${eksCluterName}`;
+
+export const awsCloudAuth = async (
```
GGonryun, line 110: Since this is AWS specific it should live in `common/aws` somewhere.

src/commands/kubeconfig.ts
```diff
@@ -0,0 +1,248 @@
...
+
+  const clusterConfig = await getAndValidateK8sIntegration(authn, args.cluster);
+
+  const { clusterId, awsAccountId, awsClusterArn } = clusterConfig;
+  const { config } = await getAwsConfig(authn, awsAccountId);
+  const { login } = config;
+
+  // Verify that the AWS auth type is supported before issuing the requests
+  if (!login?.type || login?.type === "iam") {
+    throw "This AWS account is not configured for kubectl access via the P0 CLI.\nYou can request access to the cluster using the `p0 request k8s` command.";
```
GGonryun, line 93: Shouldn't this error tell the user to login using an accepted `p0 login` method?

src/commands/kubeconfig.ts
```diff
@@ -0,0 +1,248 @@
...
+  const awsAuth = await awsCloudAuth(
+    authn,
+    awsAccountId,
+    request.generated,
+    loginType
+  );
+
+  const profile = profileName(clusterId);
+
+  // The `aws eks update-kubeconfig` command can't handle the ARN of the EKS cluster.
```
GGonryun, line 109: 🤦 amazon pls.

src/commands/kubeconfig.ts
```diff
@@ -0,0 +1,248 @@
...
+    profile,
+  ];
+
+  try {
+    // Federated access especially sometimes takes some time to propagate, so
+    // retry for up to 20 seconds just in case it takes a while.
+    const awsResult = await retryWithSleep(
+      async () => await exec("aws", updateKubeconfigArgs, { check: true }),
+      () => true,
+      2,
```
GGonryun, line 134: I feel like we should retry more often instead of once every 10 seconds for a total of two attempts, could we do something like 8 retries total, with a frequency of 2.5 seconds? If access propagates earlier we don't have to wait so long.

src/commands/kubeconfig.ts
```diff
@@ -0,0 +1,248 @@
...
+      "- <kind> / <namespace> / <name>\n" +
+      "- <kind> / <name>"
+    );
+  }
+};
+
+const extractClusterNameAndRegion = (clusterArn: string) => {
+  const INVALID_ARN_MSG = `Invalid EKS cluster ARN: ${clusterArn}`;
+  // Example EKS cluster ARN: arn:aws:eks:us-west-2:123456789012:cluster/my-testing-cluster
+  const parts = clusterArn.split(":");
```
GGonryun, line 228: ARNs come in different formats, should we use a function like https://github.com/p0-security/app/blob/main/backend/src/integrations/resources/aws/utils.ts#L83

or are we guaranteed to always get this format?

src/common/install.ts
```diff
@@ -0,0 +1,160 @@
...
+  platform: SupportedPlatform,
+  item: T,
+  installData: U
+) => {
+  const { label, commands } = installData[item];
+  print2(`To install ${label}, run the following commands:\n`);
+  for (const command of commands[platform]) {
+    print1(`  ${command}`);
+  }
+  print1(""); // Newline is useful for reading command output in a script, so send to /fd/1
```
GGonryun, line 93: Why print1 and not print2 here?

src/plugins/kubeconfig/install.ts
```diff
@@ -0,0 +1,70 @@
...
+        // needed; if kubectl is already installed, doesn't really matter how it was installed
+        // or whether it's an officially-supported architecture.
+        return kubectlInstallCommandsDarwin();
+      },
+    },
+  },
+};
+
+export const ensureEksInstall = async () => {
+  return await ensureInstall(EksItems, EksInstall);
```
gergas3, line 69: nice!

src/plugins/kubeconfig/index.ts
```diff
@@ -0,0 +1,138 @@
...
+): Promise<EksClusterConfig> => {
+  const configDoc = await getDoc<K8sConfig, object>(
+    doc(`o/${authn.identity.org.tenantId}/integrations/k8s`)
+  );
+
+  const clusterConfig = configDoc
+    .data()
+    ?.workflows.items.find(
+      (c) => c.clusterId === clusterId && c.state === "installed"
+    );
```
gergas3, line 44: Figuring out if a user can access a cluster involves evaluating the entire routing rule config. This could be a baseline check but can we rely on the backend to evaluate this as part of the `p0 request` command? In that case we don't have to also do it here.

gergas3 (commented):
My main question is whether we should make this part of the regular `p0 request` command. I think this is different from `ssh` where it is the actual executable we are substituting - equivalent would be to implement `kubectl` for Kubernetes. Instead, we are only doing post-request configuration - and that can happen after a regular request.

src/util.ts
```diff
@@ -96,3 +96,6 @@ export const assertNever = (value: never) => {
 
 export const unexpectedValueError = (value: any) =>
   new Error(`Unexpected code state: value ${value} had unexpected type`);
+
+export const stricmp = (a: string, b: string) =>
```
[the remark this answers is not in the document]
gergas3: I had a similar question: what does it actually do and is it different from lower-casing?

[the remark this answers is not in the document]
p0-andrewa: `stricmp` is a shorthand for case-insensitive string comparison from the C++ standard library. It's a common shorthand I've seen broadly in my experience, though open to suggestions if it's less familiar to others. I was just hoping to use something less verbose than, say, `caseInsensitiveCompare()` or `equalsIgnoreCase()`. Maybe `ciEquals()` as I've seen a few Stack Overflow answers use?

But this is just that, a locale-safe case-insensitive string comparison. Forcing a string to upper- or lower-case is not locale-safe and can produce unexpected results for non-ASCII characters. See [here](https://stackoverflow.com/a/2140723), for example.

src/plugins/kubeconfig/types.ts
```diff
@@ -0,0 +1,74 @@
...
+export type BaseK8sClusterConfig = {
+  clusterId: string;
+
+  clusterServer: string;
+  clusterCertificate: string;
+  state: string;
+} & (KubernetesProxyComponentConfig | KubernetesPublicComponentConfig);
+
+// k8s clusters that are not EKS clusters will not have these set...
+export type K8sClusterConfig = BaseK8sClusterConfig & {
```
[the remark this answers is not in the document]
p0-andrewa: Hmm, is this actually a problem though? `EksClusterConfig` is just a more restrictive version of `K8sClusterConfig` (where `awsAccountId` and `awsClusterArn` are not allowed to be undefined); I don't think it's an issue to, for example, allow a `EksClusterConfig` object to be provided where a `K8sClusterConfig` is expected, and therefore I see no reason to not allow them to be used interchangeably.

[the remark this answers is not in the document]
p0-andrewa: Though, to be clear, I'm not sure my approach here is the best one. I know I could have done

```typescript
export type EksClusterConfig = {
  [K in keyof K8sClusterConfig]-?: K8sClusterConfig[K];
} 
```

But I think this might be overkill, since there might be future fields added to `K8sClusterConfig` that `EksClusterConfig` need not care about.

src/plugins/kubeconfig/install.ts
```diff
@@ -0,0 +1,70 @@
...
+    case "x64": // macOS, Linux, and Windows
+      return "amd64";
+    case "arm64": // macOS and Linux only
+      return arch;
+    default:
+      throw `Unsupported system architecture for kubectl: ${arch}. Please install kubectl manually, or check that it is available in your PATH.`;
+  }
+};
+
+const kubectlInstallCommandsDarwin = (): Readonly<string[]> => {
```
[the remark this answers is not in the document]
p0-andrewa: We not only give these commands for the AWS CLU and AWS Session Manager Plugin currently (see [here](https://github.com/p0-security/p0cli/blob/2ae3841b9ae920cbee4f0dd895964270af85ab31/src/plugins/aws/ssm/install.ts)), but we attempt to automatically perform the install if possible too. This just replicates that behavior, but for `kubectl`. Indeed, we can actually give better instructions than the official website can, because system architecture information is not conveyed in User-Agent information or in HTTP Client Hints. Indeed, visiting the [official instructions for installing `kubectl` on macOS](https://kubernetes.io/docs/tasks/tools/install-kubectl-macos/) shows the Intel instructions by default, even though I'm on an Apple Silicon machine.

src/commands/aws/files.ts
p0-andrewa: `aws_security_token` is not part of the [documented spec of the AWS CLI configuration file format](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-settings), and appears to have been deprecated in 2014 and no longer appears in any AWS documentation, aside from isolated callouts in the context of backwards compatibility. Indeed, AWS' [Ansible docs](https://docs.ansible.com/ansible/latest/collections/amazon/aws/sts_assume_role_module.html#parameters) call out that,

> Support for the `EC2_SECRET_KEY` and `AWS_SECURITY_TOKEN` environment variables has been deprecated and will be removed in a release after 2024-12-01.

I think it's safe to not write out this value.

src/plugins/kubeconfig/index.ts
p0-andrewa: Technically yes, but this feels to me as more closely tied to Kubernetes as it is expecting arguments in a form that probably doesn't generalize easily without further work (i.e., I think moving this into `common/aws` might be an over-optimization at this time). Note that `assumeRoleWithIdc()` and `assumeRoleWithOktaSaml()` themselves do live within `/src/plugins/aws` and `/src/plugins/okta`, respectively.

src/commands/kubeconfig.ts
p0-andrewa: This is checking for the configured login type for the AWS account the EKS cluster is configured under in P0, so I don't think `p0 login` is relevant. Compare to the [equivalent behavior in AWS SSH](https://github.com/p0-security/p0cli/blob/2ae3841b9ae920cbee4f0dd895964270af85ab31/src/plugins/aws/ssh.ts#L64).

src/common/install.ts
p0-andrewa: This is copied from the [original implementation](https://github.com/p0-security/p0cli/blob/2ae3841b9ae920cbee4f0dd895964270af85ab31/src/plugins/aws/ssm/install.ts) that I basically just moved here. Based on the comment, I presume the intention was to send to stdout so that a harness script or something could handle installation of the tool instead of the P0 CLI itself, but I can only presume alas.

src/plugins/kubeconfig/index.ts
p0-andrewa: Oh, the purpose of this is just to check if a given k8s integration exists; it's not actually evaluating routing rules despite the `workflows` name. This just seems to be how the k8s integrations are stored, in a list under `workflows.items`:
CleanShot 2024-08-19 at 18 53 26@2x

src/commands/kubeconfig.ts
```diff
@@ -0,0 +1,248 @@
...
+    "--region",
+    clusterRegion,
+    "--profile",
+    profile,
+  ];
+
+  try {
+    // Federated access especially sometimes takes some time to propagate, so
+    // retry for up to 20 seconds just in case it takes a while.
+    const awsResult = await retryWithSleep(
```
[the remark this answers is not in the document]
p0-andrewa: Hmm, it might be? But it seems to have a lot of SSH-specific behavior, so I'd want to make a more general version to apply here. But it also seems like it might be a bit overkill in this instance as well? `accessPropagationGuard()` looks to be doing quite a bit of checking for specific, SSH-related errors, when really the only issue here is that sometimes access is denied while the permissions propagate, especially in the case of Okta SAML.

src/commands/kubeconfig.ts
```diff
@@ -0,0 +1,248 @@
...
+
+export type KubeconfigCommandArgs = {
+  cluster: string;
+  role: string;
+  resource?: string;
+  reason?: string;
+  requestedDuration?: string;
+};
+
+export const kubeconfigCommand = (yargs: yargs.Argv) =>
```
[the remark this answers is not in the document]
p0-andrewa: So, the reason I went with this approach is because it seemed to me that `p0 request` was intended to be a fairly thin wrapper around the P0 API, sharing basically the same code path as Slack and MS Teams. Some of the things you mention, like mapping args for `p0 kubeconfig` to the actual request args, I think could actually be beneficial—for example, providing a more stable interface that we could keep consistent even if we have to update the request arguments. Plus, here I'm doing some extra validation before hitting the backend for things like the requested role; would we still want things like that folded into `p0 request`? But I also see your point around `p0 kubeconfig` not being obvious, though I feel like the way `p0 request` is used is not entirely obvious either.

But yeah, let's discuss this tomorrow for sure.

src/util.ts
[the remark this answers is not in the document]
p0-andrewa: Renamed the function to `ciEquals()` and added jsdoc.

src/commands/kubeconfig.ts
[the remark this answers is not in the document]
p0-andrewa: Just to follow-up from the discussion earlier: We'll keep this as a separate `p0 kubeconfig` command for now and solicit customer feedback to determine if there's another name that makes more sense and/or if this should just be merged with the `p0 request k8s resource` command.

gergas3 (approved):

src/plugins/kubeconfig/types.ts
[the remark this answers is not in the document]
GGonryun: > Though, to be clear, I'm not sure my approach here is the best one. I know I could have done
> 
> ```ts
> export type EksClusterConfig = {
>   [K in keyof K8sClusterConfig]-?: K8sClusterConfig[K];
> } 
> ```
> 
> But I think this might be overkill, since there might be future fields added to `K8sClusterConfig` that `EksClusterConfig` need not care about.

You can use the built in typescript `Required<T>` to make optional parameters non-optional or `Partial<T>` to make no-optional parameters required and it achieves the same.

src/commands/aws/files.ts
GGonryun: In practice @gergas3 and I found that this value is still being used in practice. When debugging SSH with Venkat we found that Okta sets the SECURITY_TOKEN value which broke our integration.

See description in: https://github.com/p0-security/p0cli/commit/a21a7b261624559c2b72275bc9841fc22b0cff72

GGonryun (approved):

src/commands/aws/files.ts
p0-andrewa: Hmm, so in looking into this further, I don't think the `AWS_SECURITY_TOKEN` environment variable will be an issue because we aren't setting any environment variables; we write the credentials to the AWS CLI configuration file in a profile and then pass that profile name on to subsequent commands, and that profile is what `kubectl` is using under the hood to get the EKS token. I've confirmed this with testing.

On the other hand, in the process of testing this, I discovered a separate potential issue—for some reason, even though we invoke `aws eks update-kubeconfig` with the profile passed in as a CLI argument (e.g., `--profile profile-name`), the kubectl configuration it writes passes the profile name itself as an environment variable (`AWS_PROFILE`). This means that if the `AWS_ACCESS_KEY_ID` environment variable is set (and only that one, because that's how AWS determines if there are credentials in environment variables) it will attempt to use environment variables instead of the profile. I'm going to push one more change to address this in a quick way, which is just to detect if the `AWS_ACCESS_KEY_ID` envvar is set and emit a warning if so.

PR #97, opened 2024-08-26
K8s Generic Install

varunalla:
Updates the K8s Integration Installation document check.

### DIFF
--- a/src/plugins/kubeconfig/types.ts
+++ b/src/plugins/kubeconfig/types.ts
@@ -16,7 +16,7 @@ export type K8sClusterConfig = {
   clusterCertificate: string;
   isProxy: boolean;
   token: string;
-  publicJwk: string;
+  publicJwk?: string; // only present for proxy installs
   provider:
     | { type: "aws"; clusterArn: string; accountId: string }
     | { type: "email" };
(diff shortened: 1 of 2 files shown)


nbrahms (approved):
Lgtm

src/plugins/kubeconfig/types.ts
```diff
@@ -10,34 +10,21 @@ You should have received a copy of the GNU General Public License along with @p0
...
-export type EksClusterConfig = K8sClusterConfig & {
-  awsAccountId: string;
-  awsClusterArn: string;
-};
-
-type KubernetesProxyComponentConfig = {
-  isProxy: true;
+  isProxy: boolean;
+  token: string;
   publicJwk: string;
```
p0-andrewa, line 19: `publicJwk` is not required if `isProxy` is false; if we're not discriminating on `isProxy`, should this be optional?

p0-andrewa (commented):
One small potential issue, otherwise LGTM

src/plugins/kubeconfig/types.ts
varunalla: we could mark it as optional.

p0-andrewa (approved):

PR #99, opened 2024-09-09
Automatic gcloud CLI installation

gergas3:
Add a new `ensureInstall` method to the SSH provider type. Move the AWS SSM installation to this function, and add a new implementation for GCP that installs the gcloud CLI.

The `ensureInstall` method must be called before any other functions on the SSH provider because those functions may already use the required dependencies. In the case of GCP, the `toCliRequest` function calls `gcloud auth print-access-token`.  Add a new `prepareRequest` function is to establish the correct ordering of calls.

https://github.com/user-attachments/assets/ed96f7f4-cc34-4563-9683-c43cb21900c1

Also: logging improvement in access propagation guard: log first attempt, and log whether it's a pre-test (used only by gcp) or actual ssh session login.

### DIFF
--- a/src/commands/shared/ssh.ts
+++ b/src/commands/shared/ssh.ts
@@ -143,8 +143,28 @@ export const provisionRequest = async (
     throw "Public key mismatch. Please revoke the request and try again.";
   }
 
-  return { request: provisionedRequest, publicKey, privateKey };
+  return { provisionedRequest, publicKey, privateKey };
 };
 
-export const requestToSsh = (request: Request<PluginSshRequest>): SshRequest =>
-  SSH_PROVIDERS[request.permission.spec.type].requestToSsh(request as any);
+export const prepareRequest = async (
+  authn: Authn,
+  args: yargs.ArgumentsCamelCase<BaseSshCommandArgs>,
+  destination: string
+) => {
+  const result = await provisionRequest(authn, args, destination);
+  if (!result) {
+    throw "Server did not return a request id. Please contact support@p0.dev for assistance.";
+  }
+
+  const { provisionedRequest } = result;
+
+  const sshProvider = SSH_PROVIDERS[provisionedRequest.permission.spec.type];
+  await sshProvider.ensureInstall();
+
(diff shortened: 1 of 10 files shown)


src/plugins/ssh/index.ts
```diff
@@ -182,6 +182,17 @@ async function spawnSshNode(
 ): Promise<number | null> {
   return new Promise((resolve, reject) => {
     const provider = SSH_PROVIDERS[options.provider];
+
+    const attemptsRemaining = options.attemptsRemaining;
+    if (options.debug) {
```
GGonryun, line 187: Why did this move? Do we always want to print this out even if access already propagated (probably unlikely that it would propagate immediately)? 🤔

src/commands/shared/ssh.ts
```diff
@@ -143,12 +142,28 @@ export const provisionRequest = async (
...
+};
+
+export const prepareRequest = async (
+  authn: Authn,
+  args: yargs.ArgumentsCamelCase<BaseSshCommandArgs>,
+  destination: string
+) => {
+  const result = await provisionRequest(authn, args, destination);
+  if (!result) {
+    throw "Server did not return a request id. Please contact support@p0.dev for assistance.";
```
GGonryun, line 155: Would users know what this means? What if we used something more generic like: `Server failed to process your request. Please contact ...`

GGonryun (approved):
LGTM, only a few questions remain.

fabgo (approved):

src/plugins/ssh/index.ts
gergas3: It's only printed in debug mode. I found it useful to capture the first attempt if for some reason we never get to the second.

src/commands/shared/ssh.ts
gergas3: Good point. Maybe we should have an error code? It is useful for us to know the underlying issue when they contact us.
In the backend and frontend we have an error boundary that prints an error ID they can share with us. Filed ENG-2498

PR #100, opened 2024-09-10
ENG-2493 Add SSH -o and -R options

fabgo:
Update P0 CLI to include SSH -o (options) and -R (remote port forwarding) options

Manually verified that P0 SSH now works to enable remote port forwarding for dev instance on both AWS and GCP.

### DIFF
--- a/src/commands/ssh.ts
+++ b/src/commands/ssh.ts
@@ -63,7 +63,7 @@ export const sshCommand = (yargs: yargs.Argv) =>
         .option("o", {
           type: "string",
           describe:
-            "Can be used to give options in the format used in the configuration file.",
+            "Can be used to give options in the format used in the SSH configuration file.",
         })
         // Match `p0 request --reason`
         .option("reason", {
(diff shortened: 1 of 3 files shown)


GGonryun (commented):
Did this work for AWS too 🤔

src/commands/ssh.ts
```diff
@@ -54,6 +60,11 @@ export const sshCommand = (yargs: yargs.Argv) =>
           describe:
             "Enables forwarding of connections from an authentication agent such as ssh-agent",
         })
+        .option("o", {
+          type: "string",
+          describe:
+            "Can be used to give options in the format used in the configuration file.",
```
nbrahms, line 66:
```suggestion
            "Can be used to give options in the format used in an SSH configuration file. See `man ssh_config`.",
```

nbrahms (approved):
Per @GGonryun 's comment, I'm assuming you tested this on both GCP and AWS.

PR #106, opened 2024-09-13
ENG-2319 Specify the path to the keyfile when invoking ssh and scp.

fabgo:
Specify the path to the keyfile when invoking ssh and scp.

This is to avoid the "Too many authentication failures" that may happen when SSH has to iterate over too many keys before finding a match when attempting to connect.

See the ticket for additional details.

### DIFF
--- a/src/plugins/ssh/index.ts
+++ b/src/plugins/ssh/index.ts
@@ -243,8 +243,10 @@ const createCommand = (
 ) => {
   const commonArgs = [
     ...(args.debug ? ["-v"] : []),
+    // Explicitly specify which private key to use to avoid "Too many authentication failures"
+    // error caused by SSH trying every available key
     "-i",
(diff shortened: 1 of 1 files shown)


src/plugins/ssh/index.ts
```diff
@@ -242,6 +243,8 @@ const createCommand = (
 ) => {
   const commonArgs = [
     ...(args.debug ? ["-v"] : []),
+    "-i",
+    PRIVATE_KEY_PATH, // ENG-2319 Explicitly specify which private key to use
```
gergas3, line 247: This is a mega-nit but it may look like this is a TODO if the issue number is in the comment?
```suggestion
    PRIVATE_KEY_PATH, // Explicitly specify which private key to use to avoid "Too many authentication failures" caused by trying every key
```

gergas3 (approved):
Nice! Quicker fix than spacing out retries.

src/plugins/ssh/index.ts

PR #108, opened 2024-09-16
ENG-2498 Add support for Sentry to handle unexpected errors

fabgo:
DRAFT

### DIFF
--- a/src/commands/ssh.ts
+++ b/src/commands/ssh.ts
@@ -8,6 +8,7 @@ This file is part of @p0security/cli
 
 You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
 **/
+import { errorBoundary } from "../common/error";
 import { authenticate } from "../drivers/auth";
 import { guard } from "../drivers/firestore";
 import { sshOrScp } from "../plugins/ssh";
@@ -95,20 +96,24 @@ export const sshCommand = (yargs: yargs.Argv) =>
  * - AWS EC2 via SSM with Okta SAML
  */
 const sshAction = async (args: yargs.ArgumentsCamelCase<SshCommandArgs>) => {
-  // Prefix is required because the backend uses it to determine that this is an AWS request
-  const authn = await authenticate();
+  try {
+    // Prefix is required because the backend uses it to determine that this is an AWS request
+    const authn = await authenticate();
 
-  const { request, privateKey, sshProvider } = await prepareRequest(
-    authn,
-    args,
-    args.destination
-  );
+    const { request, privateKey, sshProvider } = await prepareRequest(
+      authn,
+      args,
+      args.destination
+    );
 
-  await sshOrScp({
-    authn,
-    request,
-    cmdArgs: args,
-    privateKey,
-    sshProvider,
-  });
+    await sshOrScp({
+      authn,
+      request,
+      cmdArgs: args,
+      privateKey,
+      sshProvider,
+    });
+  } catch (error) {
+    throw errorBoundary(error, {}, args.debug);
+  }
 };
--- a/src/common/error.ts
+++ b/src/common/error.ts
@@ -1,0 +1,47 @@
+/** Copyright © 2024-present P0 Security
+
+This file is part of @p0security/cli
+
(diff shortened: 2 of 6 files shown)


src/common/sentry.ts
```diff
@@ -0,0 +1,24 @@
...
+
+@p0security/cli is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 3 of the License.
+
+@p0security/cli is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
+**/
+import * as Sentry from "@sentry/node";
+
+const ENABLED_ENVIRONMENTS = ["stage", "staging", "prod", "production"];
```
fabgo, line 13: I do have a question about how to best handle the environment in the CLI.

src/commands/ssh.ts
```diff
@@ -95,20 +96,24 @@ export const sshCommand = (yargs: yargs.Argv) =>
...
-  });
+    await sshOrScp({
+      authn,
+      request,
+      cmdArgs: args,
+      privateKey,
+      sshProvider,
+    });
+  } catch (error) {
+    throw errorBoundary(error, {}, args.debug);
```
fabgo, line 117: Currently added boundary only for SSH command. Once everything looks good, will extend to other commands.

src/common/error.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Logs an error to container logs and Sentry with a unique identifier.
+ *
+ * @returns A generic error message that should be served to the API caller.
+ */
+export const errorBoundary = (
+  error: any,
+  logContext: Record<string, any>,
+  debug?: boolean
+) => {
+  const errorId = randomUUID();
```
fabgo, line 29: I copied this logic from app. Sentry will generate an event id. And I am wondering why we generate our own?

src/common/error.ts
```diff
@@ -0,0 +1,47 @@
...
+  };
+
+  // TODO: Convert expected errors to a subtype of Error
+  if (typeof error === "string") {
+    return error;
+  } else {
+    if (debug) {
+      print2(error);
+    }
+    Sentry.captureException(error, { extra: errorContext, level: "warning" });
```
fabgo, line 44: Should we log as warning or error?

src/common/error.ts
gergas3: We didn't know probably, or, perhaps we didn't want to rely on Sentry if this event ID generation requires the Sentry API call to succeed.

src/common/error.ts
gergas3: Only unknown / unhandeled errors will hit this code path? That's an "error" level I think.

src/common/sentry.ts
gergas3: We could still use the NODE_ENV environment variable but default to production.
We already have a bunch of configuration in `drivers/env.ts`. We could add 
```
environment: env.NODE_ENV ?? "production",
```
to that (does it have to be "prod" or "production"?) and configure our repos in with NODE_ENV=development in a `.env` file.

src/common/sentry.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const environment = process.env.NODE_ENV ?? "development";
+
+if (ENABLED_ENVIRONMENTS.includes(environment)) {
+  Sentry.init({
+    dsn: "https://4b278d15b1764983a29fb0d918cc9550@o4504696266883072.ingest.sentry.io/4504696268849152",
```
gergas3, line 19: Looks like we use a different DSN for backend and frontend. So this DSN should be a third value for the CLI.

src/common/error.ts
nbrahms: There's no need to generate this error ID IMO. In the P0 application, we do this to ensure that we don't leak any internal information to the user. But this is already running on the user's environment, and we won't have access to anything that the user already does not.

PR #110, opened 2024-09-18
ENG-2490 Support any SSH option via --

fabgo:
Support specifying any SSH option via --, e.g.:

p0 ssh dev-instance-fabian-joya --provider gcloud -- -NR '*:8080:localhost:8088' -o 'GatewayPorts yes'

Verified that both SSH and SCP work as expected, with and without --.

### DIFF
--- a/src/plugins/ssh/index.ts
+++ b/src/plugins/ssh/index.ts
@@ -240,24 +240,13 @@ const createCommand = (
   args: CommandArgs,
   proxyCommand: string[]
 ) => {
-  const commonArgs = [
-    ...(args.debug ? ["-v"] : []),
-    // Explicitly specify which private key to use to avoid "Too many authentication failures"
-    // error caused by SSH trying every available key
-    "-i",
-    PRIVATE_KEY_PATH,
-    // Only use the authentication identity specified by -i above
-    "-o",
-    "IdentitiesOnly=yes",
-    "-o",
-    `ProxyCommand=${proxyCommand.join(" ")}`,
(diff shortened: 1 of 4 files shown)


src/plugins/ssh/index.ts
```diff
@@ -292,6 +271,78 @@ const createCommand = (
...
+  }
+};
+
+const addScpArgs = (args: CommandArgs) => {
+  const sshOptions = args.sshOptions ? args.sshOptions : [];
+
+  // if a response is not received after three 5 minute attempts,
+  // the connection will be closed.
+  const serverAliveCountMaxOptionExists = sshOptions.some(
+    (opt, idx) =>
```
GGonryun, line 323: tiny nit: why not `i` 🤔 ?

src/plugins/ssh/index.ts
```diff
@@ -292,6 +271,78 @@ const createCommand = (
   };
 };
 
+/** Add common args used by both SSH & SCP to args.sshOptions.
+ *
+ * These common args are only added if they have not been explicitly specified by the end user.
+ */
+const addCommonArgs = (args: CommandArgs, proxyCommand: string[]) => {
+  const sshOptions = args.sshOptions ? args.sshOptions : [];
```
GGonryun, line 279: could we return `const sshOptions` so we can avoid having to repeat `args.sshOptions ? args.sshOptions : []`?

GGonryun (approved):

PR #112, opened 2024-09-19
ENG-2414 Add retry delay between SSH commands

fabgo:
Adds a delay between retries of the SSH command, to avoid "Exceeded Max Startups" and other connection errors:

debug1: kex_exchange_identification: banner line 0: Exceeded MaxStartups

### DIFF
--- a/src/plugins/ssh/index.ts
+++ b/src/plugins/ssh/index.ts
@@ -217,14 +217,15 @@ async function spawnSshNode(
           return;
         }
 
-        delay(RETRY_DELAY_MS).then(() =>
-          spawnSshNode({
-            ...options,
-            attemptsRemaining: attemptsRemaining - 1,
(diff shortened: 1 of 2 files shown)


src/plugins/ssh/index.ts
```diff
@@ -51,6 +51,8 @@ const SUDO_MESSAGE = /Sorry, user .+ may not run sudo on .+/; // The output of `
  */
 const DEFAULT_VALIDATION_WINDOW_MS = 5e3;
 
+const RETRY_DELAY_MS = 3000;
```
gergas3, line 54:
```suggestion
const RETRY_DELAY_MS = 500;
```

Idea being that we avoid continuous retries (If one attempt is really quick this is enough) but we are still fast in detecting access once it's provisioned

PR #114, opened 2024-09-19
ENG-2635 Replace cast of SSH options with String conversion

fabgo:


nbrahms (approved):
lgtm

gergas3:
@fabgo how about a minor release version for this fix?

PR #115, opened 2024-09-25
ENG-2677 Better error-handling of expired token

fabgo:
Improved the error-handling if 'gcloud auth print-access-token' does not return a valid access token.

### DIFF
--- a/src/plugins/google/ssh-key.ts
+++ b/src/plugins/google/ssh-key.ts
@@ -59,7 +59,15 @@ export const importSshKey = async (
   });
 
   if (!response.ok) {
-    throw `Import of SSH public key failed. HTTP error ${response.status}: ${await response.text()}`;
+    if (debug) {
+      print2(`HTTP error ${response.status}: ${await response.text()}`);
+    }
+
+    if (response.status === 401) {
+      throw `Authentication failed. Please login to Google Cloud CLI with 'gcloud auth login'`;
+    } else {
+      throw `Import of SSH public key failed.`;
+    }
   }
 
   const data: ImportSshPublicKeyResponse = await response.json();


src/plugins/google/ssh-key.ts
```diff
@@ -59,7 +59,15 @@ export const importSshKey = async (
   });
 
   if (!response.ok) {
-    throw `Import of SSH public key failed. HTTP error ${response.status}: ${await response.text()}`;
+    if (debug) {
+      print2(`HTTP error ${response.status}: ${await response.text()}`);
+    }
+
+    if (response.status === 401) {
```
gergas3, line 66: How do we get this? If the token expired we cannot even `gcloud auth print-access-token` which happens before we try to import the public key.

Were you able to observe this 401 status code returned from the importSshPublicKey endpoint?

fabgo: @gergas3  See the ticket for details.

https://linear.app/p0-security/issue/ENG-2677/p0-ssh-to-a-gcp-instance-fails-if-auth-needed#comment-0bd5cd87

It's definitely a 401, but I am not sure exactly what print-access-tokens returns as I am unable to repro.

gergas3: Oh I see, 401 is in the ticket.

`gcloud auth print-access-token` just prints the access token of the currently logged in configuration. You can run in the terminal. But it's probably an offline command. It may just read your local cred stored in your home dir, and that's why it succeeds.

image

gergas3 (approved):

src/plugins/google/ssh-key.ts
fabgo: The question is, when and if it fails, what does print-access-token return? Does it return an expired token, an error message, or something else? Hard to say without being able to reproduce it. Especially as logging is suppressed in order to protect the security of the access token :-)

PR #116, opened 2024-09-26
ENG-2685 Increase the retry delay between SSH invocations

fabgo:
This is an attempt to resolve the "MaxStartups exceeded" error which happens when too many failed authentication attempts are made to establish an SSH connection.

Testing notes:

I was able to repro this issue before increasing the delay, and not after.


gergas3 (approved):
I see, the 1s was too short!

PR #117, opened 2024-09-28
ENG-2660 Add a message once SSH connection is established

fabgo:
Add a message once an SSH connection is established, if the -N option (no remote command) is used.

### DIFF
--- a/src/plugins/ssh/index.ts
+++ b/src/plugins/ssh/index.ts
@@ -160,7 +160,7 @@ async function spawnSshNode(
         clearTimeout(timeout);
       }
       timeout = setTimeout(() => {
-        console.log("SSH connection established.");
+        print2("SSH connection established.");
       }, STDERR_TIMEOUT_PERIOD);
     };
 
--- a/src/plugins/ssh/index.ts
+++ b/src/plugins/ssh/index.ts
@@ -15,13 +15,7 @@ import { Authn } from "../../types/identity";
 import { SshProvider, SshRequest, SupportedSshProvider } from "../../types/ssh";
 import { delay } from "../../util";
 import { AwsCredentials } from "../aws/types";
-import {
-  ChildProcessByStdio,
-  StdioNull,
-  StdioPipe,
-  spawn,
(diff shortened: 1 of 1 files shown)


src/plugins/ssh/index.ts
```diff
@@ -203,11 +145,60 @@ async function spawnSshNode(
...
+      ) {
+        isEphemeralAccessDeniedException = true;
+      }
+
+      const googleLoginMatch = chunkString.match(GOOGLE_LOGIN_MESSAGE);
+      isGoogleLoginException = isGoogleLoginException || !!googleLoginMatch; // once true, always true
+      if (isGoogleLoginException) {
+        // always overwrite to false so we don't retry the access
+        isEphemeralAccessDeniedException = false;
+      }
```
nbrahms, line 193: Can we keep this GCP-specific logic within the GCP-specific SSH provider? We should ensure separation of concerns, and keep anything CSP-specific within that provider's plugin.

I know this is just a code move, but we should clean up as we go along.

src/plugins/ssh/index.ts
```diff
@@ -92,60 +88,6 @@ const UNPROVISIONED_ACCESS_MESSAGES = [
...
- * This function checks the subprocess output to see if any of the error messages
- * are printed to the error output within the first 5 seconds of startup.
- * If they are, the returned `isAccessPropagated()` function will return false.
- * When this occurs, the consumer of this function should retry the `aws` / `gcloud` command.
- *
- * Note that this function requires interception of the subprocess stderr stream.
- * This works because AWS SSM wraps the session in a single-stream pty, so we
- * do not capture stderr emitted from the wrapped shell session.
- */
-const accessPropagationGuard = (
```
nbrahms, line 111: What's the motivation to remove the sub-functioning for this PR? IMO the code is significantly easier to follow when it's more modular.

nbrahms (commented):
IMO let's keep the previous code organization. We also need to ensure GCP is kept separate from the generic logic, which we can do either here or in its own PR.

PR #118, opened 2024-09-30
ENG-2660 Always print stderr output from SSH

fabgo:
Modifies the CLI to always print stderr output from SSH child process.


fabgo:
I assume the domain allow list is set by installer? One small doubt I have is how user-friendly it is to ask customers to model a regex instead of just providing a domain list. Maybe regex should be an option, and not mandatory?

PR #120, opened 2024-09-30
ENG-2660: Print selected SSH output to stderr

fabgo:
* If debug is not enabled, it will now still print selected SSH output to stderr:
  - "Authenticated to" ... message
  - "port forwarding failed" messages
  - "Sorry, user {user} may not run sudo..." messages (on pre-test only)
* Print an error message when SSH pre-test fails.

Examples:

./p0 ssh dev-instance-fabian-joya --provider gcloud -- -NR '*:8080:localhost:8088' -o 'GatewayPorts yes'
Authenticated to dev-instance-fabian-joya (via proxy) using "publickey".
Warning: remote port forwarding failed for listen port 8080

./p0 ssh dev-instance-fabian-joya --sudo        
Failed to establish an SSH session. Use the --debug option to see additional details.

### DIFF
--- a/src/plugins/ssh/index.ts
+++ b/src/plugins/ssh/index.ts
@@ -92,6 +92,10 @@ const UNPROVISIONED_ACCESS_MESSAGES = [
   { pattern: DESTINATION_READ_ERROR },
 ];
 
+// Flags to ensure we only print each message once
+let authMessagePrinted = false;
+let portForwardingMessagePrinted = false;
+
 /** Checks if access has propagated through AWS to the SSM agent
  *
  * AWS takes about 8 minutes, GCP takes under 1 minute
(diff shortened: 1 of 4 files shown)


src/plugins/ssh/index.ts
```diff
@@ -146,6 +145,36 @@ const accessPropagationGuard = (
...
+ * @param chunkString the chunk to print
+ * @param debug true if debug output is enabled
+ */
+const parseAndPrintSshOutputToStderr = (
+  chunkString: string,
+  debug?: boolean
+) => {
+  const lines = chunkString.split("\n");
+
+  // SSH only prints the "Authenticated to" message if the verbose option (-v) is enabled
```
gergas3, line 162: So the user has to pass --debug in order to get the `Authenticated to` message?
Then do we have to always pass -v to ssh even the user didn't pass --debug? 
Do we already force -v somewhere (since it's not in this PR)?

src/plugins/ssh/index.ts
```diff
@@ -146,6 +145,36 @@ const accessPropagationGuard = (
...
+  for (const line of lines) {
+    if (debug) {
+      print2(line);
+    } else {
+      if (line.includes("Authenticated to")) {
```
gergas3, line 167: I suppose capturing this message is important to detect when port-forwarding is established. Can you add a code comment about that please?

gergas3 (commented):
> // SSH only prints the "Authenticated to" message if the verbose option (-v) is enabled

I had a question about this comment

src/plugins/ssh/index.ts
fabgo: It's here in this PR. I'll make the message more clear.

src/plugins/ssh/index.ts
```diff
@@ -315,7 +344,7 @@ const addCommonArgs = (args: CommandArgs, proxyCommand: string[]) => {
   }
 
   const verboseOptionExists = sshOptions.some((opt) => opt === "-v");
-  if (!verboseOptionExists && args.debug) {
+  if (!verboseOptionExists) {
```
fabgo, line 347: @gergas3 Here we always set -v

gergas3 (approved):

src/plugins/ssh/index.ts

PR #121, opened 2024-10-01
Throw error message if user has no email from firebase

MichaelDimitras:
If there are multiple users with the same email in Identity Platform for a particular tenant, an authentication attempt to firebase will return a user with an empty email.

On the frontend we display an error message when that happens:
https://github.com/p0-security/app/blob/5237e1a54597dc63258e56566353ad09ecd3e187/frontend/src/components/Login/hook.ts#L301-L306

But on the CLI we had no such check.  This PR adds the same check as on the frontend.  Now when a user logs in, if they have multiple identities in Identity Platform they will see the message `Can not sign in: this user has previously signed in with a different identity provider.\nPlease contact support@p0.dev to enable this user.`

https://github.com/user-attachments/assets/7d5f94f5-199f-4ec1-96d2-ef73d1802787

### DIFF
--- a/src/commands/__tests__/login.test.ts
+++ b/src/commands/__tests__/login.test.ts
@@ -23,6 +23,7 @@ jest.mock("../../drivers/auth", () => ({
 jest.mock("../../drivers/stdio");
 jest.mock("../../plugins/login");
 
+const mockSignInWithCredential = signInWithCredential as jest.Mock;
 const mockReadFile = readFile as jest.Mock;
 const mockWriteFile = writeFile as jest.Mock;
 
@@ -51,6 +52,14 @@ describe("login", () => {
     mockWriteFile.mockImplementation(async (_path, data) => {
       credentialData = data;
     });
+    mockSignInWithCredential.mockImplementation(
+      async (_auth, _firebaseCredential) =>
+        Promise.resolve({
+          user: {
+            email: "user@p0.dev",
+          },
+        })
+    );
     beforeEach(() => {
       credentialData = "";
       jest.clearAllMocks();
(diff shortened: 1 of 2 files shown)


MichaelDimitras:
The ticket has the additional goal:
>Ideally, our system will automatically update the user profile in identity platform.

I will take a stab at that as a follow-up, now that I have some understanding of the system.

src/commands/__tests__/login.test.ts
```diff
@@ -72,5 +81,14 @@ describe("login", () => {
       await login({ org: "test-org" });
       expect((signInWithCredential as jest.Mock).mock.calls).toMatchSnapshot();
     });
+    it("returns an error message if firebase cannot determine the user's email", async () => {
+      mockSignInWithCredential.mockResolvedValueOnce({
+        user: {},
+      });
+      await expect(login({ org: "test-org" })).rejects.toMatchInlineSnapshot(`
+"Can not sign in: this user has previously signed in with a different identity provider.
```
fabgo, line 89: This test is a little brittle, as any change to the message will cause the test to fail and require update. Is it not sufficient to verify that an exception is thrown?

fabgo (approved):

src/commands/__tests__/login.test.ts
MichaelDimitras: I'm trying to follow what we were already doing in this file.

```
await expect(login({ org: "test-org" })).rejects.toMatchInlineSnapshot(
    `"Could not find organization"`
);

await expect(login({ org: "test-org" })).rejects.toMatchInlineSnapshot(
    `"Unsupported login for your organization"`
    `"Unsupported login for your organization"`
);
```

Personally I like expecting on a particular message to ensure we are displaying the correct message to the user. Theoretically a logic change could expose an error with a different message.

nbrahms: I think it's fine. It's only a `yarn jest -u` to update snapshots.

PR #122, opened 2024-10-01
refactor: Separate CSP-specific functionality from generic SSH code

nbrahms:
For maintainability, anything CSP-specific for SSH should be contained within the CSP's SSH provider object.

I also sorted the provider properties.

### DIFF
--- a/src/plugins/google/ssh.ts
+++ b/src/plugins/google/ssh.ts
@@ -20,35 +20,75 @@ import { GcpSshPermissionSpec, GcpSshRequest } from "./types";
  */
 const MAX_SSH_RETRIES = 120;
 
+/**
+ * There are 7 cases of unprovisioned access in Google Cloud.
+ * These are all potentially subject to propagation delays.
+ * 1. The linux user name is not present in the user's Google Workspace profile `posixAccounts` attribute
+ * 2. The public key is not present in the user's Google Workspace profile `sshPublicKeys` attribute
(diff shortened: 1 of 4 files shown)


src/plugins/google/ssh.ts
```diff
@@ -20,35 +20,75 @@ import { GcpSshPermissionSpec, GcpSshRequest } from "./types";
...
+ * 4: results in UNAUTHORIZED_TUNNEL_USER_MESSAGE and also CONNECTION_CLOSED_MESSAGE
+ * 5a: results in UNAUTHORIZED_INSTANCES_GET_MESSAGE
+ * 6: results in SUDO_MESSAGE
+ * 7: results in DESTINATION_READ_ERROR and also CONNECTION_CLOSED_MESSAGE
+ */
+const unprovisionedAccessPatterns = [
+  { pattern: /Permission denied \(publickey\)/ },
+  {
+    // The output of `sudo -v` when the user is not allowed to run sudo
+    pattern: /Sorry, user .+ may not run sudo on .+/,
```
fabgo, line 46: Useful to keep the existing constants I think, as they may be used elsewhere. I am making a change which relies on this string for example.

fabgo (approved):

src/plugins/google/ssh.ts
nbrahms: Let's introduce that in the PR that actually reuses it?

PR #123, opened 2024-10-01
Print spinners during backend interaction

nbrahms:
Currently, we provide no user feedback whilst we contact the P0 backend. This makes it difficult for the user to distinguish between waiting on a backend response and a hung process.

Use a yarn-style Unicode spinner to indicate that the process is still live.

https://github.com/user-attachments/assets/bd58acc0-7d1a-4058-a41a-d95c3c13a929

### DIFF
--- a/src/drivers/stdio.ts
+++ b/src/drivers/stdio.ts
@@ -16,7 +16,7 @@ You should have received a copy of the GNU General Public License along with @p0
  * - Later redirection / duplication
  */
 import { sleep } from "../util";
-import { mapValues } from "lodash";
+import { Ansi, AnsiSgr } from "./ansi";
 
 /** Used to output machine-readable text to stdout
  *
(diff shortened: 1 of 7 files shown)


komal-dhull (approved):
looks good, thanks!

GGonryun:
This is coooooool! 🥳

GGonryun (approved):

src/drivers/stdio.ts
```diff
@@ -39,7 +40,50 @@ export function print2(message: any) {
...
+const Spin = {
+  items: ["⠇", "⠋", "⠙", "⠸", "⠴", "⠦"],
+  delayMs: 200,
+};
+
+/** Prints a Unicode spinner until a promise resolves */
+export const spinUntil = async <T>(message: string, promise: Promise<T>) => {
+  let isDone = false;
+  let ix = 0;
+  void promise.finally(() => (isDone = true)).catch(() => {});
```
fabgo, line 73: Should we swallow errors here?

fabgo (approved):

src/drivers/stdio.ts
nbrahms: The errors are thrown on line 77

PR #124, opened 2024-10-01
ENG-2740 Use elapsed time for SSH retry timeout

fabgo:
Use elapsed time to determine the SSH retry timeout instead of the number of retries.

### DIFF
--- a/src/plugins/aws/ssh.ts
+++ b/src/plugins/aws/ssh.ts
@@ -113,15 +113,17 @@ export const awsSshProvider: SshProvider<
 
   requestToSsh: (request) => {
     const { permission, generated } = request;
-    const { instanceId, accountId, region } = permission.spec;
-    const { idc, ssh, name } = generated;
+    const { awsResourcePermission, instanceId, accountId, region } =
+      permission.spec;
+    const { idcId, idcRegion } = awsResourcePermission.permission;
+    const { ssh, name } = generated;
     const { linuxUserName } = ssh;
     const common = { linuxUserName, accountId, region, id: instanceId };
-    return !idc
+    return !idcId || !idcRegion
       ? { ...common, role: name, type: "aws", access: "role" }
       : {
           ...common,
-          idc,
+          idc: { id: idcId, region: idcRegion },
           permissionSet: name,
           type: "aws",
           access: "idc",
--- a/src/plugins/aws/ssh.ts
+++ b/src/plugins/aws/ssh.ts
@@ -30,12 +30,89 @@ const TIME_LIMIT_MS = 10 * 60 * 1000;
 /** The name of the SessionManager port forwarding document. This document is managed by AWS.  */
 const START_SSH_SESSION_DOCUMENT_NAME = "AWS-StartSSHSession";
 
+/**There are 2 cases of unprovisioned access in AWS
+ * 1. SSM:StartSession action is missing either on the SSM document (AWS-StartSSHSession) or the EC2 instance
+ * 2. Temporary error when issuing an SCP command
+ *
+ * 1: results in UNAUTHORIZED_START_SESSION_MESSAGE
+ * 2: results in CONNECTION_CLOSED_MESSAGE
+ */
+const unprovisionedAccessPatterns = [
+  /** Matches the error message that AWS SSM prints when access is not propagated */
+  // Note that the resource will randomly be either the SSM document or the EC2 instance
+  {
+    pattern:
+      /An error occurred \(AccessDeniedException\) when calling the StartSession operation: User: arn:aws:sts::.*:assumed-role\/P0GrantsRole.* is not authorized to perform: ssm:StartSession on resource: arn:aws:.*:.*:.* because no identity-based policy allows the ssm:StartSession action/,
+  },
+  /**
+   * Matches the following error messages that AWS SSM pints when ssh authorized
+   * key access hasn't propagated to the instance yet.
+   * - Connection closed by UNKNOWN port 65535
+   * - scp: Connection closed
+   * - kex_exchange_identification: Connection closed by remote host
+   */
+  {
+    pattern: /\bConnection closed\b.*\b(?:by UNKNOWN port \d+|by remote host)?/,
+  },
+] as const;
+
 export const awsSshProvider: SshProvider<
   AwsSshPermissionSpec,
   undefined,
   AwsSshRequest,
   AwsCredentials
 > = {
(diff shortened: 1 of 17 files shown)


src/plugins/aws/ssh.ts
```diff
@@ -22,11 +22,10 @@ import {
   AwsSshRoleRequest,
 } from "./types";
 
-/** Maximum number of attempts to start an SSH session
- *
- * Each attempt consumes ~ 1 s.
+/**
+ * It can take up to 1 minute for access to propagate on AWS, so set the time limit to 10 minutes.
```
nbrahms, line 26: What's the motivation to change from 30 seconds to 10 minutes?

I'd be surprised if any end user is willing to wait 10 minutes.

src/plugins/aws/ssh.ts
```diff
@@ -22,11 +22,10 @@ import {
...
  */
-const MAX_SSH_RETRIES = 30;
+const TIME_LIMIT_MS = 10 * 60 * 1000;
```
nbrahms, line 28:
```suggestion
const TIME_LIMIT_MS = 60 * 1000;
```

Seems already long enough.

When I analyzed this previously, 99% + of provisioning finished within 8 seconds. Has that changed?

src/plugins/google/ssh.ts
```diff
@@ -14,11 +14,10 @@ import { ensureGcpSshInstall } from "./install";
 import { importSshKey } from "./ssh-key";
 import { GcpSshPermissionSpec, GcpSshRequest } from "./types";
 
-/** Maximum number of attempts to start an SSH session
- *
- * The length of each attempt varies based on the type of error from a few seconds to < 1s
+/**
+ * It typically takes < 1 minute for access to propagate on GCP, so set the time limit to 2 minutes.
```
nbrahms, line 18: Would be good to understand the CDF of successful attempts by time.

src/plugins/ssh/index.ts
```diff
@@ -185,14 +185,11 @@ async function spawnSshNode(
...
 
-    const attemptsRemaining = options.attemptsRemaining;
     if (options.debug) {
       const gerund = options.isAccessPropagationPreTest
         ? "Pre-testing"
         : "Trying";
-      print2(
-        `Waiting for access to propagate. ${gerund} SSH session... (remaining attempts: ${attemptsRemaining})`
-      );
+      print2(`Waiting for access to propagate. ${gerund} SSH session...)`);
```
nbrahms, line 192: Seems like we should display
```suggestion
      const remainingS = ((endTime - Date.now()) / 1e3 ).toFixed(1)
      print2(`Waiting for access to propagate. ${gerund} SSH session... (will wait up to ${remainingS} seconds)`);
```

nbrahms (commented):
My biggest concern is why we changed from a 30 second to 10 minute (!) timeout for AWS access.

src/types/ssh.ts
```diff
@@ -80,6 +80,8 @@ export type SshProvider<
 
   /** Unwraps this provider's types */
   requestToSsh: (request: CliPermissionSpec<PR, O>) => SR;
+  
+  timeLimit: number;
```
nbrahms, line 84: This should have some jsdoc describing what it does. I might also suggest giving it a more descriptive name:

```suggestion
  /** Amount of time, in ms, to wait between granting access and giving up on attempting an SSH connection */
  propagationTimeoutMs: number;
```

fabgo: Done

src/plugins/ssh/index.ts
fabgo: Done

src/plugins/aws/ssh.ts
fabgo: The timeout is based on existing comments in ssh/index.ts:

AWS takes about 8 minutes, GCP takes under 1 minute to fully resolve access after it is granted.
During this time, calls to `aws ssm start-session` / `gcloud compute start-iap-tunnel`
will fail randomly with an various error messages.

See also this PR, which also mentions 10 minutes:
https://github.com/p0-security/p0cli/pull/19

fabgo: @gergas3 @nbrahms I can change the timeout if you want. Just let me know what to set it to.

gergas3: Was it a typo in https://github.com/p0-security/p0cli/pull/19 then? I didn't observe such a long propagation time for AWS. 30 seconds makes more sense.

GCP takes significantly longer, 1 minute seems about right but it's my guess would be it's not >99th percentile. We continue with 1m imo.

src/plugins/ssh/index.ts
```diff
@@ -398,6 +394,6 @@ export const sshOrScp = async (args: {
     stdio: ["inherit", "inherit", "pipe"],
     debug: cmdArgs.debug,
     provider: request.type,
-    attemptsRemaining: sshProvider.maxRetries,
+    endTime: Date.now() + sshProvider.propagationTimeoutMs,
```
gergas3, line 397: To globally wait max propagationTimeoutMs, we would have to the endTime once before calling `preTestAccessPropagationIfNeeded` (above), and pass the same endTime to `spawnSshNode` (here).

Now for a GCP sudo access the total propagationTimeoutMs is 2m * 2.

gergas3 (commented):
With this change does this check in `plugins/ssh/index.ts`:
```
    if (
      match &&
      Date.now() <=
        beforeStart + (match.validationWindowMs || DEFAULT_VALIDATION_WINDOW_MS)
    ) {
      isEphemeralAccessDeniedException = true;
    }
```
still make sense? Do we want to have a timeout on individual subprocesses and consider the error non-ephemeral if we read it off the stderr too late.

I think it can be removed or set to a larger value because we have an overall cap on the time.

src/plugins/aws/ssh.ts
fabgo: Changed it to 30 seconds.

src/plugins/ssh/index.ts

src/plugins/aws/ssh.ts
nbrahms: The full info is that it takes AWS 30 minutes to get to 100% success rate. But we don't need to wait this long. We only need to wait until the chance that a single SSH attempt succeeds is high.

I'm not sure what we've changed in terms of delays, as those will affect this. But, when we were hammering it repeatedly in a loop, that took about 12 seconds.

gergas3 (approved):

PR #125, opened 2024-10-02
SSH: Update to work with newest IDC request format

nbrahms:
The P0 backend moved IDC data from the `generated` field on the request object to the `permission` field.

Update CLI to match this format.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,8 +65,8 @@
     "eslint-plugin-promise": "^6.1.1",
     "jest": "^29.7.0",
     "prettier": "^3.2.4",
-    "ts-node": "^10.9.2",
-    "ts-jest": "^29.1.2"
+    "ts-jest": "^29.1.2",
+    "ts-node": "^10.9.2"
   },
   "scripts": {
     "build": "tsc && cp -r public dist/",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -65,8 +65,8 @@
     "eslint-plugin-promise": "^6.1.1",
     "jest": "^29.7.0",
     "prettier": "^3.2.4",
-    "ts-node": "^10.9.2",
-    "ts-jest": "^29.1.2"
+    "ts-jest": "^29.1.2",
```
nbrahms, line 68: this is just yarn auto sorting its package json

src/commands/__tests__/ssh.test.ts
```diff
@@ -84,9 +85,9 @@ describe("ssh", () => {
     });
 
     it("should call p0 request with reason arg", async () => {
-      void sshCommand(yargs()).parse(
-        `ssh some-instance --reason reason --provider aws`
-      );
+      void sshCommand(yargs())
+        .fail(noop)
```
GGonryun, line 89: no action here but we can also use `false` for `.fail()` to get the no-op behavior.

> Providing false as a value for fn can be used to prevent yargs from exiting and printing a failure message.

GGonryun (approved):

src/commands/__tests__/ssh.test.ts
nbrahms: O.o

PR #126, opened 2024-10-02
Clear identity cache when a user re-authenticates with p0

GGonryun:
This PR clears the identity cache on reauth, this helps prevent stale identities from being assumed. This typically causes problems when requesting access because the new user is granted access while the old user identity is used to make the SSH connection which fails.


nbrahms (approved):
lgtm

gergas3 (approved):

PR #127, opened 2024-10-02
ENG-2768 Print an error message if SSH fails to establish a connection

fabgo:
Prints an error message if SSH fails to establish a connection.

### DIFF
--- a/src/plugins/aws/ssh.ts
+++ b/src/plugins/aws/ssh.ts
@@ -88,6 +88,8 @@ export const awsSshProvider: SshProvider<
 
   preTestAccessPropagationArgs: () => undefined,
 
+  // Starts an SSM session in the terminal by spawning `aws ssm` as a subprocess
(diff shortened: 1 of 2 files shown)


src/plugins/aws/ssh.ts
```diff
@@ -88,6 +88,8 @@ export const awsSshProvider: SshProvider<
 
   preTestAccessPropagationArgs: () => undefined,
 
+  // Starts an SSM session in the terminal by spawning `aws ssm` as a subprocess
+  // Requires `aws ssm` to be installed on the client machine.
```
gergas3, line 92: This is not accurate anymore. We aren't spawning an aws ssm subprocess. We are passing aws ssm in a ProxyCommand. I think this comment can be removed.

gergas3 (approved):

src/plugins/aws/ssh.ts
fabgo: Removed

PR #129, opened 2024-10-08
[ENG-2790] Fix broken p0 kubeconfig command

p0-andrewa:
This PR unbreaks the `p0 kubeconfig` command. Some values and structures got changed around in the k8s integration config and responses from the backend:
- Required AWS IDC metadata is no longer found in the `eksGenerated` object; it's now in `awsResourcePermission`
- AWS account ID is no longer stored separately in the k8s integration configuration, and now is extracted from the EKS cluster ARN
- `provider` was renamed `hosting` in the k8s integration configuration

It also adds some minor quality-of-live improvements to the `p0 kubeconfig` command, such as utilizing `spinUntil()` for some of the longer `await`s, and adds a more robust and reusable simple AWS ARN parser.

Added a unit test for the ARN parser, and tested everything locally in both AWS IDC and AWS Federated (Okta SAML) configurations.

### DIFF
--- a/src/commands/kubeconfig.ts
+++ b/src/commands/kubeconfig.ts
@@ -12,7 +12,8 @@ import { retryWithSleep } from "../common/retry";
 import { AnsiSgr } from "../drivers/ansi";
 import { authenticate } from "../drivers/auth";
 import { guard } from "../drivers/firestore";
-import { print2 } from "../drivers/stdio";
+import { print2, spinUntil } from "../drivers/stdio";
+import { parseArn } from "../plugins/aws/utils";
 import {
   awsCloudAuth,
   profileName,
@@ -88,12 +89,13 @@ const kubeconfigAction = async (
     throw "Required dependencies are missing; please try again after installing them, or check that they are available on the PATH.";
   }
 
+  // No spinUntil(); there is one inside requestAccessToCluster() if needed
   const request = await requestAccessToCluster(authn, args, clusterId, role);
 
   const awsAuth = await awsCloudAuth(
     authn,
     awsAccountId,
-    request.generated,
+    request,
     awsLoginType
   );
 
@@ -121,11 +123,14 @@ const kubeconfigAction = async (
   try {
     // Federated access especially sometimes takes some time to propagate, so
     // retry for up to 20 seconds just in case it takes a while.
-    const awsResult = await retryWithSleep(
-      async () => await exec("aws", updateKubeconfigArgs, { check: true }),
-      () => true,
-      8,
-      2500
+    const awsResult = await spinUntil(
+      "Waiting for AWS resources to be provisioned and updating kubeconfig for EKS",
+      retryWithSleep(
+        async () => await exec("aws", updateKubeconfigArgs, { check: true }),
+        () => true,
+        8,
+        2500
+      )
     );
     print2(awsResult.stdout);
   } catch (error: any) {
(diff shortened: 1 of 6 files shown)


src/plugins/kubeconfig/index.ts
```diff
@@ -109,14 +111,12 @@ export const requestAccessToCluster = async (
   if (!response) {
     throw "Did not receive access ID from server";
   }
-  const { id, isPreexisting } = response;
-  if (!isPreexisting) {
-    print2(
-      "Waiting for access to be provisioned. This may take up to a minute."
-    );
-  }
```
gergas3, line 117: Why can the isPreexisting be removed?

gergas3 (approved):
🙏 Thank you for the fix.
lgtm with one question

src/commands/kubeconfig.ts
```diff
@@ -233,22 +238,17 @@ const validateResourceArg = (resource: string): void => {
...
-
-  if (parts.length < 6 || !parts[3] || !parts[5]) {
-    throw INVALID_ARN_MSG;
-  }
-
-  const clusterRegion = parts[3];
-  const resource = parts[5].split("/");
+  const arn = parseArn(clusterArn);
+  const { region: clusterRegion, resource: resourceStr } = arn;
+  const resource = resourceStr.split("/");
```
nbrahms, line 243:
```suggestion
  const [resourceType, clusterName] = resourceStr.split("/");
```

🤷

src/plugins/aws/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+    throw INVALID_ARN_MSG;
+  }
+
+  return {
+    partition,
+    service,
+    region,
+    accountId,
+    resource,
+  };
```
nbrahms, line 68: With
```ts
const ARN_PATTERN =
  /^arn:(aws|aws-us-gov|aws-cn|aws-iso|aws-iso-b):([^:]*):([^:]*):([^:]*):(.*)$/;
```

```suggestion
  const invalidArnMessaage = `Invalid AWS ARN: ${arn}`;
  const match = arn.match(ARN_PATTERN);
  
  if (!match) {
    throw INVALID_ARN_MSG;
  }
  
  const [partition, service, region, accountId, resource] = match;

  return {
    partition,
    service,
    region,
    accountId,
    resource,
  };
```

src/plugins/kubeconfig/index.ts
nbrahms: Isn't `isPreexisting` about approval, not provisioning? Seems like you want to print a "waiting for approval" message when requesting access only if `!preExisting`, but always print a "waiting for ... to be provisioned" whenever an access request is made.

nbrahms (approved):
See comments

src/plugins/kubeconfig/index.ts
p0-andrewa: Yes, I believe it's about the approval, not provisioning per se, so the original message wasn't really right either. Either way, the real reason for my removing this is that it was intended as a way to tell the user that the CLI isn't frozen, just that it's waiting for the request to complete, but this is now taken care of by the `spinUntil()` call below this.

src/plugins/aws/utils.ts
p0-andrewa: This still shows the same `TS2322: Type string | undefined is not assignable to type string` error; I've adopted the change, but it looks like the `if (partition === undefined ...` block is still needed.

p0-andrewa: Okay, this is working now with a small modification, but a question before I merge—do we actually want to enforce a value for the partition in the regex? AWS doesn't make any guarantees about what partitions exist, and as long as the ARN is correctly formatted, I'm not sure it makes sense to enforce that it's one of these specifically.

PR #132, opened 2024-10-16
ENG-2884 Add multi-host support

fabgo:
Adds multi-host support to the CLI by optionally looking up the CLI configuration, including the backend host, from Firestore. This allows us to specify separate backend hosts for each tenant.

- Adds an optional config to the /orgs/{orgId} document
- Upon login, If the config is found, the config in stored locally
- When executing a command, the local config is used, if found


src/drivers/firestore.ts
```diff
@@ -8,35 +8,77 @@ This file is part of @p0security/cli
...
   DocumentReference,
   getFirestore,
   terminate,
+  Firestore,
 } from "firebase/firestore";
 
-// Your web app's Firebase configuration
-const firebaseConfig = config.fs;
+let firestore: Firestore;
+let auth: Auth;
```
gergas3, line 33: I think we should avoid mutable global variables like these. Esp if they are about authentication. We may make a mistake in updating them and then the user is attempting to log in to the wrong host (= P0 environment).

An idea here is to keep the module-level `const app, firestore, auth` objects and only initialize another firestore when we access the public `orgs` document. That would be a different method, not `getDoc`, but e.g. `getPublicDoc` and it creates its own firestore using the bootstrap config.

src/drivers/config.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Config } from "../types/org";
+import { P0_PATH } from "../util";
+import { bootstrapConfig } from "./env";
+import { print2 } from "./stdio";
+import fs from "fs/promises";
+import path from "path";
+
+export const CONFIG_FILE_PATH = path.join(P0_PATH, "config.json");
+
+export let tenantConfig = bootstrapConfig;
```
gergas3, line 20: A mutable variable at the module-level will lead to problems imo. We will have the wrong config in it, e.g. someone makes a change that overwrites its value directly.

Can this be accessed via a getter?

```
let tenantConfig = bootstrapConfig;
...
export const getTenantConfig = () => tenantConfig
```

src/commands/login.ts
```diff
@@ -48,8 +56,15 @@ export const login = async (
   await clearIdentityCache();
   await writeIdentity(orgWithSlug, tokenResponse);
 
+  if (orgData.config) {
+    await saveTenantConfig(orgData.config);
+  }
```
gergas3, line 61: Doesn't this have to move up, right after `if (!orgData) throw "Could not find organization";`?

Afaict currently we will log the user in to the bootstrap project and whatever the org is configured for login there. I think for orgs that actually use a different host, the login information is going to be in the Firestore of that other host (= P0 environment), and the p0-prod document for their org will only contain a config that points to the right environment.

So we don't want to log these users in to our p0-prod. We want them to log in to their environment. When `loginFn` is called tenantConfig should already be updated with the right config.

src/commands/login.ts
```diff
@@ -32,6 +38,8 @@ export const login = async (
   args: { org: string },
   options?: { skipAuthenticate?: boolean }
 ) => {
+  initializeFirebase({ useBootstrapConfig: true });
+
```
gergas3, line 42: Seems the line below is the only `getDoc` call where we read a public collection.

Could we create a special `getPublicDoc` function that always initializes and uses the bootstrap config to retrieve the tenant config?

src/commands/login.ts
gergas3: A second thought: we could always save tenantConfig and use the bootstrapConfig as a fallback right here. Then we don't have to pre-assign in `drivers/config.ts`, tenantConfig can be undefined until we save here. 
```
await saveTenantConfig(orgData.config ?? bootstrapConfig);
```
Or, we could also read tenantConfig lazily in a `getTenantConfig` function that checks the variable, then checks the file, then loads. Since then you can check `if tenantConfig === undefined`.

src/commands/login.ts
fabgo: This is a great suggestion. Done.

src/commands/login.ts
fabgo: Done

src/drivers/firestore.ts
fabgo: I was able to move auth to a local variable. But it's a bit tricky to make everything const. Is this a concern even if these are not exported?

src/drivers/config.ts
fabgo: Done

src/drivers/firestore.ts
gergas3: imo not exported is good enough

src/commands/login.ts
```diff
@@ -49,7 +59,10 @@ export const login = async (
   await writeIdentity(orgWithSlug, tokenResponse);
 
   // validate auth
-  if (!options?.skipAuthenticate) await authenticate({ noRefresh: true });
+  if (!options?.skipAuthenticate) {
+    const identity = await loadCredentials({ noRefresh: true });
+    await authenticateToFirebase(identity);
```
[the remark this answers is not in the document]
fabgo: It's because authenticate now saves the config & initializes firebase. I would like to rename this method now, but couldn't think of a better name.

And there used to be a circular call chain, which I am trying to break:

authenticate -> loadCredentials -> login -> authenticate

src/commands/__tests__/__snapshots__/login.test.ts.snap
```diff
@@ -1,7 +1,14 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`login organization exists should write the user's identity to the file system 1`] = `
+exports[`login organization exists should write the user's identity & config to the file system 1`] = `
 [
+  [
+    "/Users/fabianjoya/.p0/config.json",
```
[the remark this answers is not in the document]
fabgo: Fixed

gergas3 (approved):

PR #133, opened 2024-10-24
Add ssh-keygen command

gergas3:
This command can be used if user only wants to create the key pair for SSH. One use case when user requests access through Slack instead of through the CLI and they need to provide the public key in Slack.

The command could be extended to force regeneration of the key pair. I did not include that option because already provisioned keys would not work anymore.


src/commands/ssh-keygen.ts
```diff
@@ -0,0 +1,22 @@
...
+import yargs from "yargs";
+
+export const sshKeyGenCommand = (yargs: yargs.Argv) =>
+  yargs.command(
+    "ssh-keygen",
+    `Generate an SSH key pair for use with P0 and store it in your ${P0_KEY_FOLDER} folder, if one does not exist already. This key pair is also automatically generated by using any "p0 ssh/scp ..." command.`,
+    (yargs) => yargs,
+    async () => {
+      await createKeyPair();
+    }
```
nbrahms, line 21: nit:
```suggestion
    createKeyPair
```

nbrahms (approved):
lgtm

src/commands/ssh-keygen.ts
gergas3: `createKeyPair` has a return value so we have to do this

PR #135, opened 2024-10-30
ENG-2960 Read ssoProvider from config host, not p0-prod

fabgo:
Reads the ssoProvider for login from the host specified in the config, instead of defaulting to p0-prod.

### DIFF
--- a/src/commands/login.ts
+++ b/src/commands/login.ts
@@ -14,8 +14,8 @@ import {
   IDENTITY_FILE_PATH,
 } from "../drivers/auth";
 import { saveConfig } from "../drivers/config";
-import { bootstrapConfig } from "../drivers/env";
-import { fsShutdownGuard, publicDoc } from "../drivers/firestore";
+import { fsShutdownGuard, initializeFirebase } from "../drivers/firestore";
+import { doc } from "../drivers/firestore";
 import { print2 } from "../drivers/stdio";
 import { pluginLoginMap } from "../plugins/login";
 import { TokenResponse } from "../types/oidc";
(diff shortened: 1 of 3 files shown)


src/commands/login.ts
```diff
@@ -34,13 +34,13 @@ export const login = async (
   args: { org: string },
   options?: { skipAuthenticate?: boolean }
 ) => {
-  const orgDoc = await getDoc<RawOrgData, object>(
-    publicDoc(`orgs/${args.org}`)
-  );
+  await saveConfig(args.org);
+  await initializeFirebase();
```
MichaelDimitras, line 38: I don't follow why we need to add `initializeFirebase` here/. Wasn't it already initialized if we were able to read the doc before?

src/drivers/config.ts
```diff
@@ -22,11 +25,19 @@ export function getTenantConfig(): Config {
   return tenantConfig;
 }
 
-export async function saveConfig(config: Config) {
+export async function saveConfig(orgId: string) {
+  const orgDoc = await getDoc<RawOrgData, object>(
+    bootstrapDoc(`orgs/${orgId}`)
+  );
```
MichaelDimitras, line 31: Should there be the null check `if (!orgData) throw` here as well?

We wouldn't want to write the config if there was no org found I assume

src/commands/login.ts
fabgo: There are two Firebase, the bootstrap one (which is configured to point to p0-prod) and the regular one, which uses whatever is specified in the config.

src/drivers/config.ts
fabgo: It's there on line 33. We use the bootstrap config if no config was found.

gergas3 (approved):
lgtm!

PR #137, opened 2024-10-31
SSH requesting access message

MichaelDimitras:
This PR makes two changes to the status messages we use for SSH requests in the CLI.

1. Replaces the "Requesting access" status message with a more generic "Checking for access in P0"
2. Adds the message "Existing access found. Connecting to node." in the scenario where a user had access approved previously and is waiting to be connected to the node.  Previously we showed nothing.

This only affects `p0 ssh <destination>` and `p0 scp <destination>`.  It does not affect any `p0 request` flows.

Before:

https://github.com/user-attachments/assets/cc0fce18-ab55-4074-af16-14a43a6ce06f

After:

https://github.com/user-attachments/assets/3c31ae41-437f-4328-a798-096710515dcc

Addresses: https://linear.app/p0-security/issue/ENG-2827/requesting-access-spinner-is-displayed-when-it-shouldnt

### DIFF
--- a/src/commands/shared/request.ts
+++ b/src/commands/shared/request.ts
@@ -100,8 +100,16 @@ export const request =
   ): Promise<RequestResponse<T> | undefined> => {
     const resolvedAuthn = authn ?? (await authenticate());
     const { userCredential } = resolvedAuthn;
+    const accessMessage = (message?: string) => {
+      switch (message) {
+        case "approval-required":
+          return "Checking for access in P0";
+        default:
+          return "Requesting access";
+      }
+    };
     const data = await spinUntil(
-      options?.accessMessage ? options.accessMessage : "Requesting access",
+      accessMessage(options?.message),
       fetchCommand<RequestResponse<T>>(resolvedAuthn, args, [
         command,
         ...args.arguments,
--- a/src/commands/shared/request.ts
+++ b/src/commands/shared/request.ts
@@ -94,13 +94,14 @@ export const request =
     }>,
     authn?: Authn,
     options?: {
+      accessMessage?: string;
       message?: "all" | "approval-required" | "none";
     }
   ): Promise<RequestResponse<T> | undefined> => {
     const resolvedAuthn = authn ?? (await authenticate());
     const { userCredential } = resolvedAuthn;
     const data = await spinUntil(
-      "Requesting access",
+      options?.accessMessage ? options.accessMessage : "Requesting access",
       fetchCommand<RequestResponse<T>>(resolvedAuthn, args, [
         command,
         ...args.arguments,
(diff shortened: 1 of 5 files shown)


src/commands/shared/request.ts
```diff
@@ -94,13 +94,14 @@ export const request =
...
     options?: {
+      accessMessage?: string;
       message?: "all" | "approval-required" | "none";
     }
   ): Promise<RequestResponse<T> | undefined> => {
     const resolvedAuthn = authn ?? (await authenticate());
     const { userCredential } = resolvedAuthn;
     const data = await spinUntil(
-      "Requesting access",
+      options?.accessMessage ? options.accessMessage : "Requesting access",
```
GGonryun, line 104: should this be dependent on the type union? `"all" | "approval-required" | "none"`

So if we see: "approval-required" we can still say: "requesting access" 🤔

MichaelDimitras: Since `"approval-required"` is hardcoded for the ssh session request, I don't think we can use it to change the message in this case. I think the request would still have `approval-required` even if the user already had been approved in the past?

https://github.com/p0-security/p0cli/pull/137/files#diff-c16738fe5df540e2a49c613e959ebaa27241c6ce8389095b737f76c65bc44f8eL125

GGonryun: I see, I see 🤔 .

GGonryun (approved):

src/commands/shared/ssh.ts
```diff
@@ -131,6 +131,7 @@ export const provisionRequest = async (
   }
   const { id, isPreexisting } = response;
   if (!isPreexisting) print2("Waiting for access to be provisioned");
+  else print2("Existing access found.  Connecting to node.");
```
gergas3, line 134: We called them `instance` in the CLI help
```suggestion
  else print2("Existing access found.  Connecting to instance.");
```

src/commands/shared/request.ts
gergas3: You can make it dependent like this:
```
const accessMessage = (message?: string) => {
  switch (message) {
    case "approval-required":
      return "Requesting access";
    default:
      return "Requesting access";
  }
};
```
then here
```suggestion
      accessMessage(options?.message),,
```

MichaelDimitras: I made this change, but the `"approval-required"` message type will return an accessMessage of `Checking for access in P0`.

This makes this change apply to `p0 kubeconfig` as well which I think is desirable since it's similar flow to starting an ssh session.

```
const accessMessage = (message?: string) => {
  switch (message) {
    case "approval-required":
     return "Checking for access in P0";;
    default:
      return "Requesting access";
  }
};
```

PR #140, opened 2024-11-13
Update existing SSH permission requests to match new backend data structures

GGonryun:
This PR changes the permission request data structures for AWS, Azure, and Google Cloud SSH to be in line with our most recent changes to the P0 backend.

These change impact the construction of SSH commands. This latest version is necessary to continue making valid SSH requests.


src/plugins/google/types.ts
```diff
@@ -12,14 +12,13 @@ import { PermissionSpec } from "../../types/request";
 import { CliPermissionSpec } from "../../types/ssh";
 import { CommonSshPermissionSpec } from "../ssh/types";
 
-export type GcpSshPermission = {
-  spec: CommonSshPermissionSpec & {
+export type GcpSshPermission = CommonSshPermissionSpec & {
+  zone: string;
+  provider: "gcloud";
+  resource: {
```
p0-andrewa, line 18: Nit: Should `provider` be the first field here? That seems to be the pattern I've observed with other discriminated types.

p0-andrewa (commented):
LGTM, just a small style nit/question.

p0-andrewa (approved):
Oops, clicked too quickly and forgot to actually select "Approve".

src/plugins/aws/ssh.ts
```diff
@@ -112,11 +112,12 @@ export const awsSshProvider: SshProvider<
...
-      permission.spec;
-    const { idcId, idcRegion } = awsResourcePermission.permission;
-    const { ssh, name } = generated;
-    const { linuxUserName } = ssh;
+    const { resource, region } = permission;
+    const { idcId, idcRegion, instanceId, accountId } = resource;
+    const {
+      linuxUserName,
+      resource: { name },
+    } = generated;
```
nbrahms, line 120: Nested destructuring gets a bit verbose and a bit complicated:

```suggestion
    const { linuxUserName, resource } = generated;
    const { name } = resource;
```

nbrahms (approved):
I'd probably re-write the PR description. I believe this change is more about updating for the new permission-request data structure used by the P0 backend.

Other than that, just the tiniest of style feedback.

src/plugins/aws/ssh.ts
GGonryun: There's a collision between `permission.resource` and `generated.resource` but i used an alias to make it work 👍

PR #141, opened 2024-11-15
ls: display number of items that were truncated 

komal-dhull:
Addresses ENG-2998. Improves CLI usability when there are a large number of results by giving the user an idea of the total number of results for their query. We request twice the requested size, and if the total number of results is within the amount, we display the number of remaining results. If not, we just display 'of many'

image
image

### DIFF
--- a/src/commands/ls.ts
+++ b/src/commands/ls.ts
@@ -13,9 +13,11 @@ import { fetchCommand } from "../drivers/api";
 import { authenticate } from "../drivers/auth";
 import { fsShutdownGuard } from "../drivers/firestore";
 import { print2, print1, spinUntil } from "../drivers/stdio";
-import { max, orderBy } from "lodash";
+import { max, orderBy, pullAt, slice } from "lodash";
 import pluralize from "pluralize";
 import yargs from "yargs";
+
+const DEFAULT_RESPONSE_SIZE = 15;
 
 type LsResponse = {
   ok: true;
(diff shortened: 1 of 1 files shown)


src/commands/ls.ts
```diff
@@ -48,15 +50,36 @@ export const lsCommand = (yargs: yargs.Argv) =>
...
     arguments: string[];
   }>
 ) => {
   const authn = await authenticate();
+  const { convertedArgs, requestedSize } = convertLsSizeArg(args.arguments);
+
   const data = await spinUntil(
     "Listing accessible resources",
-    fetchCommand<LsResponse>(authn, args, ["ls", ...args.arguments])
+    fetchCommand<LsResponse>(authn, args, ["ls", ...convertedArgs])
```
GGonryun, line 82: im a bit iffy on this approach, naively i would have expected the `fetch` command to return the total number of results before performing fuzzy search. why are we taking this "doubling" approach? 🤔

src/commands/ls.ts
```diff
@@ -78,7 +102,11 @@ const ls = async (
     print2(
       `Showing${truncationPart} ${label}${postfixPart}.\nResources labeled with * are already accessible to you:`
     );
-    const sortedItems = orderBy(data.items, "isPreexisting", "desc");
+    const sortedItems = orderBy(
+      slice(data.items, 0, requestedSize),
+      "isPreexisting",
+      "desc"
+    );
```
nbrahms, line 109: Nit: (for clarity and concision):

```suggestion
    const truncated = slice(data.items, 0, requestedSize);
    const sortedItems = orderBy(truncated, "isPreexisting", "desc");
```

nbrahms (approved):
lgtm

src/commands/ls.ts
GGonryun: appreciate the extra context. maybe in the future we can improve our current architecture to support this but im on board with these changes.

PR #142, opened 2024-11-15
update ssh command arguments to use shared properties

GGonryun:
This PR updates the ssh command argument `account` to use the new shared name: `parent`. The parent argument unifies inputs of `accountId` for AWS, `projectId` for Google Cloud, and `subscriptionId` for Azure.

image

### DIFF
--- a/src/commands/ssh.ts
+++ b/src/commands/ssh.ts
@@ -43,9 +43,9 @@ export const sshCommand = (yargs: yargs.Argv) =>
           describe: "Reason access is needed",
           type: "string",
         })
-        .option("account", {
+        .option("parent", {
           type: "string",
-          describe: "The account on which the instance is located",
+          describe: "The parent resource on which the instance is located",
         })
         .option("provider", {
           type: "string",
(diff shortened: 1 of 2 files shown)


GGonryun:
https://linear.app/p0-security/issue/ENG-3108/ssh-cli-update-the-ssh-account-property-to-shared-parent-property

src/commands/ssh.ts
```diff
@@ -43,9 +43,9 @@ export const sshCommand = (yargs: yargs.Argv) =>
           describe: "Reason access is needed",
           type: "string",
         })
-        .option("account", {
+        .option("parent", {
           type: "string",
-          describe: "The account on which the instance is located",
+          describe: "The parent resource on which the instance is located",
```
p0-andrewa, line 48: Nit: I wonder if this should be a little more descriptive, maybe by way of brief examples? Maybe something along the lines of, "e.g., account ID for AWS, resource group ID for Azure"?

p0-andrewa (approved):
LGTM, just one small thing

src/commands/ssh.ts
GGonryun: updated 👍

PR #143, opened 2024-11-21
Initial implementation of Azure SSH

p0-andrewa:
This PR implements the initial version of Azure SSH for the P0 CLI. It is fully functional and works with P0 instances configured with an Azure iam-write integration and an Azure SSH integration.

The process of establishing an Azure SSH connection, as implemented in this PR, works as follows:

1. Issue an Azure SSH request to the backend
2. Create a temporary directory to hold the SSH certificate pair generated by `az ssh cert`
3. Log in to Azure using `az login` (implemented in previous PR) and set the active subscription using `az account set --subscription`
4. Generate the temporary SSH certificate pair using `az ssh cert`
5. Spawn an Azure Network Bastion tunnel, connecting to the remote VM on port 22 and listening locally on a random ephemeral port
    - If the local port is in-use or the command otherwise fails, select a new random port number and retry up to 2 additional times
6. Generate additional SSH options needed (such as specifying the just-generated SSH certificate pair and the local port of the bastion tunnel)
    - This is new functionality for the `SshProvider` framework, allowing individual implementations to specify additional options to pass to `ssh`/`scp`
7. Spawn SSH or SCP, whichever was needed
8. When SSH or SCP terminates, teardown the tunnel by sending a SIGINT to the tunnel process, then delete the temporary directory containing the SSH certificate pair (which will expire after an hour anyway)

## Testing
Tested locally against the dev Azure environment using my local dev instance of P0 to manage requests. Note that it is no longer necessary to use the `azure-ssh-dev-base` branch as this PR now includes all necessary changes for Azure SSH to work.

Things explicitly tested to work:
- SSH to an Azure VM via Network Bastion
- SCP files to and from an Azure VM via Network Bastion
    - Resolves ENG-2559
- `--sudo` to request sudo access (but it takes some time after access is granted for `sudo` to work, even after one can access the VM; unsure why)
- SSH options, including port forwarding in both directions (`-L`/`-R`)
    - All options with the exception of `ssh -p` and `scp -P` (see below), should work since this is just invoking usual `ssh` and `scp` executables.
    - Resolves ENG-2563 and ENG-2562

Known limitations:
- One must log in to Azure again with every single `ssh`/`scp` attempt to ensure credentials are refreshed. To be addressed as part of ENG 3019.
- Custom ports for SSH (i.e., the `ssh -p` and `scp -P` options) are not currently supported. To be addressed as part of ENG 3142.
- Mostly missing debug output and no reproduction commands. To be addressed as part of ENG 3131.

## Demo videos
Part 1: Basic `ssh` and `scp` use
https://github.com/user-attachments/assets/d972d8be-71d3-4243-b67a-1dff4ba11891

Part 2: Demonstrating `sudo` access when using the `--sudo` flag
https://github.com/user-attachments/assets/5d1549db-8c3f-4cad-8fb2-522facef29d0

Part 3: Demonstrating an SSH tunnel using local port forwarding (`ssh -L`)
https://github.com/user-attachments/assets/3a2b3115-f584-44e7-895f-35a19dcea6bf

Part 4: Demonstrating an SSH tunnel using remote port forwarding (`ssh -R`)
https://github.com/user-attachments/assets/fdf32dcb-1a0b-4364-b4e4-e1d53201254c


src/plugins/azure/auth.ts
```diff
@@ -0,0 +1,18 @@
...
+
+You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
+**/
+import { exec } from "../../util";
+
+export const azLogin = async (subscriptionId: string) => {
+  await exec("az", ["login"], { check: true });
+  await exec("az", ["account", "set", "--subscription", subscriptionId], {
+    check: true,
+  });
```
gergas3, line 17: Is setting the subscription necessary? 
`az network bastion` has a `--subscription` option that we could use.
(The performance is a bit slow probably due to the many `az` subprocesses we have to call.)

src/plugins/azure/ssh.ts
```diff
@@ -45,31 +54,74 @@ export const azureSshProvider: SshProvider<
 
   propagationTimeoutMs: PROPAGATION_TIMEOUT_LIMIT_MS,
 
-  // TODO: Implement
+  // TODO: Determine if necessary
```
gergas3, line 57: If sudo takes time to propagate we can use the preTestAccessPropagationArgs (see GCP) to verify that sudo works before actually opening the session for the user.
This is just fyi for later.

src/plugins/azure/ssh.ts
```diff
@@ -45,31 +54,74 @@ export const azureSshProvider: SshProvider<
...
+
+    const { path: keyPath, cleanup: sshKeyPathCleanup } =
+      await createTempDirectoryForKeys();
+    await generateSshKeyAndAzureAdCert(keyPath);
+
+    const { killTunnel, tunnelLocalPort } =
+      await trySpawnBastionTunnel(request);
+
+    const sshPrivateKeyPath = path.join(keyPath, AD_SSH_KEY_PRIVATE);
+    const sshCertificateKeyPath = path.join(keyPath, AD_CERT_FILENAME);
```
gergas3, line 78: We'd probably want to clean up the directory if the next step fails.

```suggestion
try {
    const { killTunnel, tunnelLocalPort } =
      await trySpawnBastionTunnel(request);

    const sshPrivateKeyPath = path.join(keyPath, AD_SSH_KEY_PRIVATE);
    const sshCertificateKeyPath = path.join(keyPath, AD_CERT_FILENAME);
 } finally {
   await sshKeyPathCleanup()
 } 
```

src/plugins/ssh/index.ts
```diff
@@ -273,7 +287,7 @@ const addCommonArgs = (args: CommandArgs, proxyCommand: string[]) => {
       opt === "-o" && sshOptions[idx + 1]?.startsWith("ProxyCommand")
   );
 
-  if (!proxyCommandExists) {
+  if (!proxyCommandExists && proxyCommand.length > 0) {
```
gergas3, line 290: nit: could make the new `proxyCommand.length > 0` condition part of `proxyCommandExists`

gergas3 (commented):
👏 fantastic! So many tickets addressed in one go with this solution.

I had a couple of comments about adding try-catches that I think are worth addressing.

src/plugins/azure/auth.ts
GGonryun: That could be fixed here: https://linear.app/p0-security/issue/ENG-3129/ssh-investigate-how-az-login-works-with-cross-subscription-bastion

We're not 100% sure how cross subscription access will behave.

src/plugins/azure/tunnel.ts
```diff
@@ -0,0 +1,123 @@
...
+        "--port",
+        port,
+      ],
+      // Spawn the process in detached mode so that it is in its own process group; this lets us kill it and all
+      // descendent processes together.
+      { detached: true }
+    );
+
+    child.on("exit", (code) => {
+      if (tunnelReady) return;
```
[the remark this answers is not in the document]
p0-andrewa: This is true. I'll replace this with a `code !== 0` check, but my plan is to improve the debugging output in general as part of ENG-3131, and change to outputting the contents of `stdout` and `stderr` only when `--debug` is set.

src/plugins/azure/tunnel.ts
```diff
@@ -0,0 +1,123 @@
...
+              // Kill the process and all its descendents via killing the process group; this is only possible
+              // because we launched the process with `detached: true` above. This is necessary because `az` is
+              // actually a bash script that spawns a Python process, and we need to kill the Python process as well.
+              // SIGINT is equivalent to pressing Ctrl-C in the terminal; allows for the tunnel process to perform any
+              // necessary cleanup of its own before exiting. The negative PID is what indicates that we want to kill
+              // the whole process group.
+              process.kill(-child.pid, "SIGINT");
+            }
+
+            child.unref();
```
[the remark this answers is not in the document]
p0-andrewa: The only reason I have the `if (child.pid)` check is because `child.pid` is typed as `number | undefined`, when it being `undefined` if the process failed to exec. `child` itself should never be undefined, even if the process fails to spawn, and `unref()` should just do nothing in that case.

src/plugins/azure/tunnel.ts
```diff
@@ -0,0 +1,123 @@
...
+            processTerminated = true;
+
+            if (child.pid) {
+              // Kill the process and all its descendents via killing the process group; this is only possible
+              // because we launched the process with `detached: true` above. This is necessary because `az` is
+              // actually a bash script that spawns a Python process, and we need to kill the Python process as well.
+              // SIGINT is equivalent to pressing Ctrl-C in the terminal; allows for the tunnel process to perform any
+              // necessary cleanup of its own before exiting. The negative PID is what indicates that we want to kill
+              // the whole process group.
+              process.kill(-child.pid, "SIGINT");
```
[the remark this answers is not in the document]
p0-andrewa: I think it should be safe, but since Miguel had similar feedback during our pair review yesterday, I went ahead and added a spin-wait for the process to terminate after the `SIGINT`, followed by a `SIGKILL` if it didn't terminate within 5 seconds. I also added a `try`/`catch` around the `process.kill()` calls, and only call `child.unref()` if there was an error.

src/plugins/azure/ssh.ts
```diff
@@ -45,31 +54,74 @@ export const azureSshProvider: SshProvider<
...
-    ...request,
-    cliLocalData: {
-      linuxUserName: await importSshKey(request.permission.publicKey, options),
-    },
-  }),
+  toCliRequest: async (request) => {
+    return {
+      ...request,
+      cliLocalData: {
+        linuxUserName: request.principal,
```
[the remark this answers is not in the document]
p0-andrewa: Yup, it's really the email address, `@` character included. And it is needed to work; the `aad_certhandler` utility used for Entra ID logins to check the validity of the presented SSH certificate uses the username specified in the `ssh` command to match the certificate to an Entra ID user:

`/etc/ssh/sshd_config`:
```
Match User *@*,????????-????-????-????-????????????    # Added by aadsshlogin installer
    AuthenticationMethods publickey
    PubkeyAuthentication yes
    AuthorizedKeysCommand /usr/sbin/aad_certhandler %u %k
    AuthorizedKeysCommandUser root
```

Note that the `Match` block is looking for usernames that contain an `@`. The `az network bastion ssh` command also uses the user's email to log in. (This `Match` block is also looking for usernames that are a UUID it seems, so I wonder if that would work as a fallback if we run into issues in the future.)

And if I run `whoami` on the VM, it shows my email address as my username:

```
$ whoami   
andrew.adams@p0.dev
```

src/plugins/azure/auth.ts
p0-andrewa: There's another reason why this is necessary—to select the correct Entra ID directory. We do already provide the Bastion's subscription ID to the `az network bastion` command because we provide the whole Bastion resource ID scope string, but in my testing the command would still sometimes fail because `az login` running in non-interactive mode doesn't have the opportunity to ask you which directory to use if you have access to multiple, and if it chooses the wrong one you can get unexpected authorization errors. By calling `az account set --subscription`, we ensure the correct Entra ID directory is in use.

`az account set` reliably takes about 240ms in my testing, so I don't think it's impacting performance much in particular.

src/plugins/azure/ssh.ts
p0-andrewa: Created https://linear.app/p0-security/issue/ENG-3149/ to track this work; will also update the TODO comment.

src/plugins/azure/ssh.ts
p0-andrewa: Done, but implemented as a `try`/`catch` instead of `try`/`finally`, since we don't want to trigger the key path cleanup here if we're able to spawn the tunnel successfully.

src/plugins/ssh/index.ts
p0-andrewa: These are actually two different checks so I don't want to combine them. `proxyCommandExists` is checking whether there is a `ProxyCommand` already specified by the user as an argument to `p0 ssh` (e.g., `p0 ssh ... -- -o 'ProxyCommand /some/command`) so we don't overwrite it. I'm not sure what the use case for this behavior is, but I figure it's out of scope to touch this. `proxyCommand.length > 0` is checking whether the `SshProvider` in use specifies a `ProxyCommand`, of which the Azure provider is the only one that doesn't. Without the check, we will try to append an empty `ProxyCommand` to the SSH arguments (e.g., `-o 'ProxyCommand'`) which is an error.

What I will do though is do some renaming here to make this more clear, but let me know if you strongly think they should be combined.

src/plugins/azure/ssh.ts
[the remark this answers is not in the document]
gergas3: Azure is truly amazing. GCP removes special characters including `@`.

src/plugins/azure/ssh.ts
gergas3: Oh right, finally is wrong here!

src/plugins/ssh/index.ts
gergas3: I see, I thought `proxyCommandExists` is whether the provider has it. The new names are clear! 🙏

gergas3 (approved):

PR #144, opened 2024-11-22
Implement reproCommands for Azure SSH

p0-andrewa:
This PR improves the Azure SSH debugging experience by adding reproduction commands for Azure SSH, which are displayed when invoking `p0 ssh` with the `--debug` option. It outputs all the `az` commands, as used by the P0 CLI, that are needed to manually reproduce the SSH connection process, as well as the final `ssh`/`scp` command too. I also included some minor refactoring which both made the implementation easier, as well as removed some redundancy from the `ssh` options we end up passing for Azure SSH (e.g., previously, we technically specified two separate private keys to `ssh`, the usual P0 CLI-managed key as well as the ephemeral SSH certificate private key generated by `az ssh cert`).

The first commit also includes removing a TODO for ENG 3129, which is no longer required and replaces it with a describe comment, as a tiny "while I'm here" change.

## Testing
Tested locally. Repro commands were tested against my local P0 instance connecting to my Azure test VM (to confirm that the repro commands I added in this PR work as expected) and against an AWS EC2 instance in Federated auth mode (to confirm that I didn't break any existing repro commands with my changes).

## Screenshots
Example repro commands for `p0 ssh` for Azure SSH (note: redaction of sensitive values by my terminal, not part of the P0 CLI output):
Repro commands ssh

Example repro commands for `p0 scp` for Azure SSH:
Repro commands scp

### DIFF
--- a/src/plugins/ssh/index.ts
+++ b/src/plugins/ssh/index.ts
@@ -214,12 +214,12 @@ const createCommand = (
   setupData: SshAdditionalSetup | undefined,
   proxyCommand: string[]
 ) => {
-  addCommonArgs(args, proxyCommand);
+  addCommonArgs(args, proxyCommand, setupData);
 
-  const sshOptions = setupData?.sshOptions ?? [];
+  const sshOptionsOverrides = setupData?.sshOptions ?? [];
   const port = setupData?.port;
 
-  const argsOverride = sshOptions.flatMap((opt) => ["-o", opt]);
+  const argsOverride = sshOptionsOverrides.flatMap((opt) => ["-o", opt]);
 
   if ("source" in args) {
     addScpArgs(args);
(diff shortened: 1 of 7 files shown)


src/plugins/ssh/index.ts
```diff
@@ -278,7 +279,12 @@ const addCommonArgs = (
   // Explicitly specify which private key to use to avoid "Too many authentication failures"
   // error caused by SSH trying every available key
   if (!identityFileOptionExists) {
-    sshOptions.push("-i", PRIVATE_KEY_PATH);
+    if (setupData?.identityFile) {
+      sshOptions.push("-i", setupData.identityFile);
+    } else {
+      sshOptions.push("-i", PRIVATE_KEY_PATH);
+    }
+
```
gergas3, line 287: nit
```suggestion
    sshOptions.push("-i", setupData.identityFile ?? PRIVATE_KEY_PATH);
```

gergas3 (approved):

GGonryun (approved):

src/plugins/ssh/index.ts
p0-andrewa: d'oh

p0-andrewa:
Simple cherry-pick on main and addressing nits. Will merge once checks are done running.

PR #145, opened 2024-11-22
Add debugging output for Azure SSH provider

p0-andrewa:
This PR improves the Azure SSH debugging experience by adding debug output for the various `az` commands being run before opening the `ssh`/`scp` session. Specifically, when the `--debug` flag is set, the CLI will output the stdout and stderr for all the `az` commands it runs. Note though that it _won't_ set the `--debug` flag for the `az` commands themselves as this outputs far too much information, with the exception of the `az network bastion tunnel` command; while it still outputs far too much information in debug mode, it has an unfortunate tendency to hide certain errors unless you pass it the `--debug` flag, so I did add it there.

Some additional information is also logged by the CLI itself now when the `--debug` flag is set.

## Testing
Tested locally against my Azure VM in our test tenant, both with and without the `--debug` flag.

## Screenshots
Example output when _not_ using `--debug`; matches output from before this PR (as before, sensitive redaction value by my terminal, not the CLI):
Output without debug

Examples of output when using `--debug`:
Initial startup:
Output with debug 1

Generating Azure AD SSH certificate:
Output with debug 2

Starting tunnel (and showing repro commands added in previous PR):
Output with debug 3

Closing SSH session and shutting down tunnel:
Output with debug 4

### DIFF
--- a/src/plugins/azure/auth.ts
+++ b/src/plugins/azure/auth.ts
@@ -8,6 +8,7 @@ This file is part of @p0security/cli
 
 You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
 **/
+import { print2 } from "../../drivers/stdio";
 import { exec } from "../../util";
 
 export const azLoginCommand = () => ({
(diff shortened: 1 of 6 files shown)


GGonryun (approved):

p0-andrewa:
Simple rebase on main; will merge when checks are done running.

PR #147, opened 2024-11-25
Add sudo propagation check for azure

GGonryun:
This PR implements Azure's `preTestAccessPropagationArgs` and adds a new parameter to the `SSHProvider` type to support valid access errors. 

Sudo access to an azure node can sometimes take an unexpectedly long time to propagate fully and if a user connects before sudo access propagates then they'll need to restart their session in order to pick up the sudo access. Therefore it is necessary to validate sudo access before letting a user complete their connection. 

We get an error code of 1 instead of an error code 0 after sudo access is granted when performing the pretest for an Azure instance. This means that we needed to extend the `SSHProvider` interface a little to include potentially valid error messages and override the error code of 1.

> In the case a user doesn't have sudo access, sudo -nv will exit with code 1 and print Sorry, user ${username} may not run sudo on ${host}. to stderr. 
>
> If the user does have sudo access, sudo -nv will exit with code 1 (so, the same as the no access case) but print sudo: a password is required to stderr.

### DIFF
--- a/src/plugins/ssh/index.ts
+++ b/src/plugins/ssh/index.ts
@@ -57,17 +57,26 @@ const accessPropagationGuard = (
 ) => {
   let isEphemeralAccessDeniedException = false;
   let isLoginException = false;
+  let isValidPreTestError = false;
 
   child.stderr.on("data", (chunk) => {
     const chunkString: string = chunk.toString("utf-8");
     parseAndPrintSshOutputToStderr(chunkString, options);
 
-    const match = provider.unprovisionedAccessPatterns.find((message) =>
-      chunkString.match(message.pattern)
+    const matchUnprovisionedPattern = provider.unprovisionedAccessPatterns.find(
+      (message) => chunkString.match(message.pattern)
     );
 
-    if (match) {
+    const matchPreTestPattern = provider.validPreTestAccessPatterns?.find(
+      (message) => chunkString.match(message.pattern)
+    );
+
+    if (matchUnprovisionedPattern) {
       isEphemeralAccessDeniedException = true;
+    }
+
+    if (matchPreTestPattern && !matchUnprovisionedPattern) {
+      isValidPreTestError = true;
     }
 
     if (provider.loginRequiredPattern) {
@@ -83,6 +92,7 @@ const accessPropagationGuard = (
   return {
     isAccessPropagated: () => !isEphemeralAccessDeniedException,
     isLoginException: () => isLoginException,
+    isValidPreTestError: () => isValidPreTestError,
   };
 };
 
@@ -170,11 +180,8 @@ async function spawnSshNode(
     );
 
     // TODO ENG-2284 support login with Google Cloud: currently return a boolean to indicate if the exception was a Google login error.
-    const { isAccessPropagated, isLoginException } = accessPropagationGuard(
-      provider,
-      child,
-      options
-    );
+    const { isAccessPropagated, isLoginException, isValidPreTestError } =
+      accessPropagationGuard(provider, child, options);
 
     const exitListener = child.on("exit", (code) => {
       exitListener.unref();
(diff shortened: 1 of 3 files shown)


src/plugins/ssh/index.ts
```diff
@@ -57,19 +57,28 @@ const accessPropagationGuard = (
...
-    const match = provider.unprovisionedAccessPatterns.find((message) =>
-      chunkString.match(message.pattern)
+    const matchUnprovisionedPattern = provider.unprovisionedAccessPatterns.find(
+      (message) => chunkString.match(message.pattern)
     );
 
-    if (match) {
+    const matchPreTestPattern = provider.validPreTestAccessPatterns?.find(
+      (message) => chunkString.match(message.pattern)
+    );
```
gergas3, line 72: I wonder if `accessPropagationGuard` has to "know" about pre-testing.

There could be a new arg `validAccessPatterns` passed to `accessPropagationGuard` to make this more general. The `isAccessPropagated()` returned function can take into. account the pre-test pattern so the caller in `src/plugins/ssh/index.ts` doesn't need to know about pre-testing either.

(Also, `provider` arg could be ultimately removed by just taking patterns.)

```
const accessPropagationGuard = (
  child: ChildProcessByStdio<null, null, Readable>,
  invalidAccessPatterns: readonly AccessPattern[],
  validAccessPatterns: readonly AccessPattern[] | undefined,
  loginRequiredPattern: RegExp | undefined,
  options: SpawnSshNodeOptions
) => {
  let isEphemeralAccessDeniedException = false;
  let isLoginException = false;
  let isValidError = false;

  child.stderr.on("data", (chunk) => {
    const chunkString: string = chunk.toString("utf-8");
    parseAndPrintSshOutputToStderr(chunkString, options);

    const matchUnprovisionedPattern = invalidAccessPatterns.find((message) =>
      chunkString.match(message.pattern)
    );

    const matchPreTestPattern = validAccessPatterns?.find((message) =>
      chunkString.match(message.pattern)
    );

    if (matchUnprovisionedPattern) {
      isEphemeralAccessDeniedException = true;
    }

    if (matchPreTestPattern && !matchUnprovisionedPattern) {
      isValidError = true;
    }

    if (loginRequiredPattern) {
      const loginMatch = chunkString.match(loginRequiredPattern);
      isLoginException = isLoginException || !!loginMatch; // once true, always true
    }

    if (isLoginException) {
      isEphemeralAccessDeniedException = false; // always overwrite to false so we don't retry the access
    }
  });

  return {
    isAccessPropagated: () =>
      !isEphemeralAccessDeniedException &&
      (!validAccessPatterns || isValidError),
    isLoginException: () => isLoginException,
  };
};
```
`src/plugins/ssh/index.ts`:
```
...
    const { isAccessPropagated, isLoginException } = accessPropagationGuard(
      child,
      provider.unprovisionedAccessPatterns,
      provider.validPreTestAccessPatterns,
      provider.loginRequiredPattern,
      options
    );
...
      options.abortController?.abort(code);
      if (options.isAccessPropagationPreTest && isAccessPropagated()) {
        // override the exit code to 0 if the expected error was found, this means access is ready.
        resolve(0);
        return;
      }
      print2(`SSH session terminated`);
      resolve(code);
```

src/plugins/ssh/index.ts
```diff
@@ -427,22 +444,21 @@ export const sshOrScp = async (args: {
...
       credential,
+      setupData,
       endTime
     );
     if (exitCode && exitCode !== 0) {
       return exitCode; // Only exit if there was an error when pre-testing
     }
 
-    sshNodeExit = await spawnSshNode({
+    return await spawnSshNode({
```
gergas3, line 461: Neat!

gergas3 (approved):
I had one optional suggestion about removing the notion of pre-test from the accessPropagationGuard, and hopefully that can simplify a bit.
(Haven't tested.)

src/plugins/ssh/index.ts
GGonryun: i like it! 👍 fixed.

PR #149, opened 2024-12-05
Azure SSH: clear current identity before logging in again

GGonryun:
This PR adds runs the `az logout` before logging in so that we wipe any currently connected identities. This resolves a handful of issues we're seeing for example: 

1. granting a user access, relinquishing access, and then granting again will use an old cached identity which will have a temporary block on access.
2. granting sudo access, and regular user access and relinquishing one temporarily blocks access to the account.
3. if someone runs `az login` and uses a different identity that has the same subscription id.

### DIFF
--- a/src/plugins/azure/auth.ts
+++ b/src/plugins/azure/auth.ts
@@ -16,6 +16,11 @@ export const azLoginCommand = () => ({
   args: ["login"],
 });
 
+export const azLogoutCommand = () => ({
+  command: "az",
+  args: ["logout"],
+});
+
 export const azAccountSetCommand = (subscriptionId: string) => ({
   command: "az",
   args: ["account", "set", "--subscription", subscriptionId],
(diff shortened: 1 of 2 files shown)


src/plugins/azure/auth.ts
```diff
@@ -16,6 +16,11 @@ export const azLoginCommand = () => ({
   args: ["login"],
 });
 
+export const azLogoutCommand = () => ({
+  command: "az",
+  args: ["logout"],
+});
```
p0-andrewa, line 22: Should the `az logout` command be added to the `azureSshProvider.reproCommands()` output for completeness?

p0-andrewa (commented):
One question inline

src/plugins/azure/auth.ts
GGonryun: fixed 👍

PR #150, opened 2024-12-05
Allow ctrl+d to exit terminal when valid access patterns exist

GGonryun:
We recently introduced `validAccessPatterns` in #147 which checks to make sure a certain termination string exists when performing propagation checks. Consequentially for Azure, when we exit the terminal we're now always attempting to verify that an ephemeral access message is valid and failing the check.

Two fixes were possible, 1) we either only check `isValidError` if we're in a propagation check or 2) we default `isValidError` to true and we only set it to false if we encounter an error.

We didn't go with the first option in #147 because we want to remove the concept of `isAccessPropagationPreTest` from the `accessPropagationGuard`, but checking for valid error messages should only apply during the pretest phase. This PR instead only passes `validAccessPatterns` if we're performing a pre-test to the propagation guard.

### DIFF
--- a/src/plugins/ssh/index.ts
+++ b/src/plugins/ssh/index.ts
@@ -95,7 +95,7 @@ const accessPropagationGuard = (
     isAccessPropagated: () => {
       return (
         !isEphemeralAccessDeniedException &&
-        (!validAccessPatterns?.length || isValidError)
+        (!validAccessPatterns || isValidError)
       );
     },
     isLoginException: () => isLoginException,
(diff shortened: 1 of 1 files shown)


src/plugins/ssh/index.ts
```diff
@@ -32,11 +32,6 @@ import {
 } from "node:child_process";
 import { Readable } from "node:stream";
 
-/** Maximum amount of time after SSH subprocess starts to check for {@link UNPROVISIONED_ACCESS_MESSAGES}
- *  in the process's stderr
- */
-const DEFAULT_VALIDATION_WINDOW_MS = 5e3;
```
GGonryun, line 38: this wasn't being used anywhere

p0-andrewa (approved):
Looks good and makes sense to me, though happy to be overruled

gergas3 (approved):

PR #154, opened 2025-01-13
Azure SSH: use generated linuxUserName from backend service

GGonryun:
Only the azure principal name can be used as the linux identity. This change makes sure that we always default to the generated linuxUserName when attempting to log into an SSH node using Azure. The user principal in the permission points to the user's email in P0 which can be different than the identity they have in Azure Active Directory. 

If the user accidentally signs in with the wrong account they will get an error that the expected user name doesn't match their current identity. This makes it easier to debug `public key` and `permission denied` errors.

### DIFF
--- a/src/plugins/azure/auth.ts
+++ b/src/plugins/azure/auth.ts
@@ -116,9 +116,9 @@ const getUserPrincipalName = async ({ debug }: { debug?: boolean }) => {
     const { command, args } = azAccountShowUserPrincipalName();
     const accountShowResult = await exec(command, args, { check: true });
     if (debug) {
+      print2(`Found account information...`);
       print2(accountShowResult.stdout);
       print2(accountShowResult.stderr);
-      print2(`Found account information...`);
     }
     return accountShowResult.stdout.trim();
   } catch (error: any) {
--- a/src/plugins/azure/auth.ts
+++ b/src/plugins/azure/auth.ts
@@ -51,6 +51,11 @@ export const azAccountSetCommand = (subscriptionId: string) => ({
   args: ["account", "set", "--subscription", subscriptionId],
 });
 
+export const azAccountShowUserPrincipalName = () => ({
+  command: "az",
+  args: ["account", "show", "--query", "user.name", "-o", "tsv"],
+});
+
 const performLogout = async ({ debug }: { debug?: boolean }) => {
   try {
     const { command: azLogoutExe, args: azLogoutArgs } = azLogoutCommand();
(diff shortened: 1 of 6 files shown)


src/plugins/ssh/index.ts
```diff
@@ -264,7 +264,7 @@ const createCommand = (
       ...(args.sshOptions ? args.sshOptions : []),
       ...argsOverride,
       ...(port ? ["-p", port] : []),
-      `${data.linuxUserName}@${data.id}`,
+      `${setupData?.linuxUserName ?? data.linuxUserName}@${data.id}`,
```
p0-andrewa, line 267: Do we need to make a similar change for `scp` too? Might be a little tricky since we'll have to bring the overridden username into `replaceHostWithInstance()`.

p0-andrewa (commented):
One question inline

p0-andrewa (approved):

src/plugins/azure/auth.ts
```diff
@@ -51,6 +51,11 @@ export const azAccountSetCommand = (subscriptionId: string) => ({
   args: ["account", "set", "--subscription", subscriptionId],
 });
 
+export const azAccountShowUserPrincipalName = () => ({
+  command: "az",
+  args: ["account", "show", "--query", "user.name", "-o", "tsv"],
```
fabgo, line 56: Is this a call to our API backend? If yes, how does this correlate with your other PR. Just curious to understand how this works.

GGonryun: It's not a call to our backend. The other PR sets the linux user name as the Active Directory UserPrincipalName which is what is used whenever someone logs into an Azure Node. This PR checks what the current logged in user's UserPrincipalName is. 

The problem the other PR solves is that it also needs to check to see if a user is an external user, if so the UserPrincipalName is set as the "mail" field. While this command does that automatically.

src/plugins/azure/auth.ts
```diff
@@ -106,6 +111,21 @@ const performSetAccount = async (
...
 };
 
+const getUserPrincipalName = async ({ debug }: { debug?: boolean }) => {
+  try {
+    const { command, args } = azAccountShowUserPrincipalName();
+    const accountShowResult = await exec(command, args, { check: true });
+    if (debug) {
+      print2(accountShowResult.stdout);
+      print2(accountShowResult.stderr);
+      print2(`Found account information...`);
```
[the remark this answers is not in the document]
GGonryun: fixed 👍

PR #156, opened 2025-01-14
Make Google OAuth client parameters optional

gergas3:
The `google` object in the tenant config contains the Client ID and Client Secret.
This config is then stored in plain text in the user's home directory.
Although the Client Secret is not really a secret, storing it in plain text gives the impression that the P0 CLI is not secure.

This PR renames the `clientSecret` to `publicClientSecretForPkce` and adds more explanation in code comments. (Although these comments are not visible in the saved `config.json` file.)

The PR also makes the `google` parameters optional so this becomes a non-issue for organizations with different login methods like Okta, Ping, etc.

User-friendly error if the organization is configured with Google login but the OAuth parameters are missing:
Screenshot 2025-01-14 at 5 32 31 PM

Saved config without `google` object if the org is not configured with Google:
Screenshot 2025-01-14 at 5 35 44 PM
(The cat'ed values here are already publicly visible in the code base)

Saved config with the `google` object and renamed `publicClientSecretForPkce` if the org is configured with Google:
Screenshot 2025-01-14 at 5 38 35 PM
(The cat'ed values here are already publicly visible in the code base)

### DIFF
--- a/src/drivers/config.ts
+++ b/src/drivers/config.ts
@@ -8,7 +8,7 @@ This file is part of @p0security/cli
 
 You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
 **/
-import { Config, GoogleConfig, RawOrgData } from "../types/org";
+import { Config, RawOrgData } from "../types/org";
 import { P0_PATH } from "../util";
 import { bootstrapConfig } from "./env";
 import { bootstrapDoc } from "./firestore";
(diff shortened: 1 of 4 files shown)


src/types/org.ts
```diff
@@ -8,9 +8,27 @@ This file is part of @p0security/cli
 
 You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
 **/
-import { bootstrapConfig } from "../drivers/env";
+type NonGoogleConfig = {
```
GGonryun, line 11: nit: maybe we can call this something more descriptive like `ApplicationConfig`? calling it `NonGoogle` feels a bit strange.

src/types/org.ts
```diff
@@ -8,9 +8,27 @@ This file is part of @p0security/cli
...
+
+export type GoogleConfig = NonGoogleConfig & {
+  google: {
+    clientId: string;
+    publicClientSecretForPkce: string;
+  };
+};
 
-export type Config = typeof bootstrapConfig;
+export type Config = GoogleConfig | NonGoogleConfig;
```
fabgo, line 31: Do we not want to highlight that this is the bootstrap config? I named it bootstrapConfig because it is required in the Splunk case to download the real config from Firestore.

fabgo (approved):

src/types/org.ts
gergas3: Hm, this is a type, and it's the type of both the bootstrap and non-bootstrap configs, isn't it?
Maybe if you post suggested change it's clearer to me.

src/drivers/config.ts
```diff
@@ -25,6 +25,14 @@ export function getTenantConfig(): Config {
   return tenantConfig;
 }
 
+/** Use only if the organization is configured with Google login to P0 */
+export function getGoogleTenantConfig(): GoogleConfig {
```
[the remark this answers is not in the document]
gergas3: I replaced all functions with `const` declarations in this file.

src/types/org.ts
gergas3: Do you mean the type should be names `BootstrapConfig` instead of `Config`?
(Note the type name didn't change, it has been `Config`)

fabgo: I missed that. Ignore my comment.

GGonryun (approved):

PR #159, opened 2025-01-28
Azure SSH: Print errors during log in attempts when debugging

GGonryun:
# Overview
This PR reduces the number of login attempts for Azure. Instead of logging out and logging after every single access request, we now only attempt a login when:

1. A user has been granted access to a subscription they didn't have access to previously. 
2. A user's token issuer has become invalid.

# Understanding the different authentication failure modes

## A user has gained access to a subscription they didn't have access to previously.
This happens more frequently to guests, most orgs will have members that are assigned to a tenant/subscription which prevents this error from coming up. There's no way for us to poll for a subscription without explicitly logging out and logging in again. The user is warned that they may need to try again in a few minutes. 

```
Subscription bcf379ca-fdd9-4ac4-adae-8f33c376fc6c not found. If access was recently granted, please try again in a few minutes. If the issue persists, please contact support@p0.dev.
```

## A user's token issuer has become invalid.
In order to support the second use case above I had to add an abort controller and listen to messages from the azure bastion tunnel telling us the access is invalid. This code would be greatly simplified if instead of trying to automatically log a user in we told them to execute `az logout && az login` manually. This error is inconsistent and usually only comes up

# Removing known errors
This PR removes the `KnownError` type and it's corresponding function because it did not provide enough flexibility for processing errors.


src/plugins/azure/auth.ts
```diff
@@ -10,47 +10,37 @@ You should have received a copy of the GNU General Public License along with @p0
 **/
 import { print2 } from "../../drivers/stdio";
 import { exec } from "../../util";
-import { KnownError } from "./types";
-
-const knownLoginErrors: KnownError[] = [
```
GGonryun, line 15: This mini-framework wasn't flexible enough and just made the error handling more complex, I removed it in favor of just being explicit about which errors are being handled by each function.

src/plugins/azure/auth.ts
```diff
@@ -10,47 +10,37 @@ You should have received a copy of the GNU General Public License along with @p0
...
+  "If the issue persists, please contact support@p0.dev.";
+export const NASCENT_ACCESS_GRANT_MESSAGE =
+  "If access was recently granted, please try again in a few minutes.";
+export const ABORT_AUTHORIZATION_FAILED_MESSAGE = `Your Microsoft Token Cache is out of date. Run 'az account clear' and 'az login' to refresh your credentials. ${CONTACT_SUPPORT_MESSAGE}`;
+
+export const azLoginCommand = (tenantId: string) => ({
   command: "az",
-  args: ["login"],
+  args: [
+    "login",
```
GGonryun, line 33: This is a small stop-gap to help prevent users from picking the wrong account. When the login prompt appears it'll only show users account's that belong to that tenant.

src/plugins/azure/auth.ts
```diff
@@ -10,47 +10,37 @@ You should have received a copy of the GNU General Public License along with @p0
...
+    "--scope",
+    "https://management.core.windows.net//.default",
+    "--tenant",
+    tenantId,
+  ],
 });
 
-export const azLogoutCommand = () => ({
+export const azAccountClearCommand = () => ({
```
GGonryun, line 41: According to the official docs, we should call `az account clear` instead of `az logout` to refresh the token. The logout command can be potentially disruptive if a user is trying to access many different subscriptions at once.

https://learn.microsoft.com/en-us/cli/azure/manage-azure-subscriptions-azure-cli?tabs=bash#clear-your-subscription-cache

bgberr (approved):

src/plugins/azure/auth.ts
```diff
@@ -10,47 +10,37 @@ You should have received a copy of the GNU General Public License along with @p0
...
+  "If access was recently granted, please try again in a few minutes.";
+export const ABORT_AUTHORIZATION_FAILED_MESSAGE = `Your Microsoft Token Cache is out of date. Run 'az account clear' and 'az login' to refresh your credentials. ${CONTACT_SUPPORT_MESSAGE}`;
+
+export const azLoginCommand = (tenantId: string) => ({
   command: "az",
-  args: ["login"],
+  args: [
+    "login",
+    "--scope",
+    "https://management.core.windows.net//.default",
```
p0-andrewa, line 35: The fact that you need to have two slashes before `.default` is dumb. 😐

src/plugins/azure/auth.ts
```diff
@@ -75,46 +66,76 @@ const performLogout = async ({ debug }: { debug?: boolean }) => {
...
+    return loginResult.stdout;
   } catch (error: any) {
-    throw normalizeAzureCliError(error, knownLoginErrors, { debug });
+    if (debug) {
+      print2("Failed to log in to Azure...");
+      print2(error.stderr);
+    }
+
+    if (FAILED_TO_RESOLVE_TENANT_PATTERN.test(error.stderr)) {
+      throw `Failed to resolve tenant "${directoryId}". If access was recently granted, please try again in a few minutes. If the issue persists, please contact support@p0.dev.`;
```
p0-andrewa, line 97: Nit: Looks like you could use `CONTACT_SUPPORT_MESSAGE` and `NASCENT_ACCESS_GRANT_MESSAGE` to avoid repetition?

src/plugins/azure/auth.ts
```diff
@@ -75,46 +66,76 @@ const performLogout = async ({ debug }: { debug?: boolean }) => {
...
+      print2("Failed to set active Azure subscription...");
+      print2(error.stderr);
+    }
+
+    if (SUBSCRIPTION_NOT_FOUND_PATTERN.test(error.stderr)) {
+      await performAccountClear({ debug });
+      const output = await performLogin(request.directoryId, { debug });
+      if (!output.includes(request.subscriptionId))
+        throw `Subscription ${request.subscriptionId} not found. ${NASCENT_ACCESS_GRANT_MESSAGE}`;
+      await performSetAccount(request, { debug });
```
p0-andrewa, line 135: Is there a risk this could recurse infinitely if, for some reason, the subscription ID is returned in the login output but the `az account set` command still fails? I know this probably wouldn't ever happen, but since `az` is an external system we can't control, might be good to have an explicit recursion depth check?

p0-andrewa (commented):
One nit and one potentially more significant issue noted.

src/plugins/azure/auth.ts
GGonryun: fixed 👍

PR #160, opened 2025-01-29
Add the public P0 logo to README.md

GGonryun:


dhing1024 (approved):
The only other consideration would be how it looks on GitHub dark mode but otherwise this is good.

Screenshot 2025-01-29 at 3 11 10 PM

GGonryun:
> The only other consideration would be how it looks on GitHub dark mode but otherwise this is good.
> 



We have one with a white background, do you think we should swap to that one? 🤔

dhing1024:
> > The only other consideration would be how it looks on GitHub dark mode but otherwise this is good.
> 
> We have one with a white background, do you think we should swap to that one? 🤔

It might be a little more universal

PR #161, opened 2025-01-29
ssh-resolve and ssh-proxy

bgberr:
This is to allow users to run
`ssh <node-name>` rather than `p0 ssh <node-name>`

It adds two commands to the cli, `ssh-resolve` and `ssh-proxy` meant to be used in an ssh config such as the following
```
Match exec "p0 ssh-resolve %h -q"
 Include .p0/ssh/configs/*.config
```

`ssh-resolve` checks if the resource exists in p0 and if the user has `approved` access to it.  If so it sets up any needed keys and writes an ssh config to `.p0/ssh/configs/` as well as request information to a temporary file, and returns 0 (success). Otherwise it returns 1 (silently if `--quiet` is passed)

`ssh-proxy` is invoked as the `ProxyCommand` by `ssh` when the config file generated by `ssh-resolve` is used, and reads the files written by `ssh-resolve` (and passed to it as cli args) and runs the relevant proxy command as well as any needed setup/teardown

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -18,25 +18,6 @@
             ],
             "sourceMaps": true,
             "runtimeExecutable": "ts-node",
-        },
-        {
-            "type": "node",
-            "request": "launch",
-            "name": "Launch P0 CLI (with args)",
-            "skipFiles": [
-                "<node_internals>/**"
-            ],
-            "program": "${workspaceFolder}/src/index.ts",
-            "preLaunchTask": "tsc: build - tsconfig.json",
-            "outFiles": [
-                "${workspaceFolder}/**/*.js"
-            ],
-            "sourceMaps": true,
-            "runtimeExecutable": "ts-node",
-            "args": [
-                "ssh-resolve",
-                "miguel-node-1"
-            ]
         }
     ]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -20,4 +20,4 @@
             "runtimeExecutable": "ts-node",
         }
     ]
-} 
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -20,4 +20,4 @@
             "runtimeExecutable": "ts-node",
         }
     ]
-}
+} 
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -18,6 +18,25 @@
             ],
             "sourceMaps": true,
             "runtimeExecutable": "ts-node",
+        },
+        {
+            "type": "node",
+            "request": "launch",
+            "name": "Launch P0 CLI (with args)",
+            "skipFiles": [
+                "<node_internals>/**"
+            ],
+            "program": "${workspaceFolder}/src/index.ts",
+            "preLaunchTask": "tsc: build - tsconfig.json",
+            "outFiles": [
+                "${workspaceFolder}/**/*.js"
+            ],
+            "sourceMaps": true,
+            "runtimeExecutable": "ts-node",
+            "args": [
+                "ssh-resolve",
+                "miguel-node-1"
+            ]
         }
     ]
 }
--- a/src/commands/shared/request.ts
+++ b/src/commands/shared/request.ts
@@ -108,19 +108,17 @@ export const request =
           return "Requesting access";
       }
     };
+
+    const fetchCommandPromise = fetchCommand<RequestResponse<T>>(
+      resolvedAuthn,
+      args,
+      [command, ...args.arguments]
+    );
+
     const data =
       options?.message != "quiet"
-        ? await spinUntil(
-            accessMessage(options?.message),
-            fetchCommand<RequestResponse<T>>(resolvedAuthn, args, [
-              command,
-              ...args.arguments,
-            ])
-          )
-        : await fetchCommand<RequestResponse<T>>(resolvedAuthn, args, [
-            command,
-            ...args.arguments,
-          ]);
+        ? await spinUntil(accessMessage(options?.message), fetchCommandPromise)
+        : await fetchCommandPromise;
 
     if (data && "ok" in data && "message" in data && data.ok) {
       const logMessage =
--- a/src/commands/shared/request.ts
+++ b/src/commands/shared/request.ts
@@ -95,7 +95,7 @@ export const request =
     authn?: Authn,
     options?: {
       accessMessage?: string;
-      message?: "all" | "approval-required" | "none";
+      message?: "all" | "approval-required" | "none" | "quiet";
     }
   ): Promise<RequestResponse<T> | undefined> => {
     const resolvedAuthn = authn ?? (await authenticate());
@@ -108,13 +108,19 @@ export const request =
           return "Requesting access";
       }
     };
-    const data = await spinUntil(
-      accessMessage(options?.message),
-      fetchCommand<RequestResponse<T>>(resolvedAuthn, args, [
-        command,
-        ...args.arguments,
-      ])
-    );
+    const data =
+      options?.message != "quiet"
+        ? await spinUntil(
+            accessMessage(options?.message),
+            fetchCommand<RequestResponse<T>>(resolvedAuthn, args, [
+              command,
+              ...args.arguments,
+            ])
+          )
+        : await fetchCommand<RequestResponse<T>>(resolvedAuthn, args, [
+            command,
+            ...args.arguments,
+          ]);
 
     if (data && "ok" in data && "message" in data && data.ok) {
       const logMessage =
--- a/src/commands/shared/ssh.ts
+++ b/src/commands/shared/ssh.ts
@@ -165,7 +165,6 @@ export const provisionRequest = async (
     return;
   }
   const { id, isPreexisting } = response;
-  if (!quiet) print2("Did not receive access ID from server");
   if (!isPreexisting) print2("Waiting for access to be provisioned");
   else print2("Existing access found.  Connecting to instance.");
 
--- a/src/commands/shared/ssh.ts
+++ b/src/commands/shared/ssh.ts
@@ -52,9 +52,17 @@ export type SshCommandArgs = BaseSshCommandArgs & {
   command?: string;
 };
 
-export type SshProxyCommandArgs = BaseSshCommandArgs & {
+export type SshResolveCommandArgs = SshCommandArgs & {
+  quiet?: boolean;
+};
+
+export type SshProxyCommandArgs = {
   destination: string;
   port: string;
+  provider: "aws" | "azure" | "gcloud";
+  requestJson: string;
+  debug?: boolean;
+  identityFile: string;
 };
 
 export type CommandArgs = ScpCommandArgs | SshCommandArgs;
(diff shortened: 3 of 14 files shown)


.vscode/launch.json
```diff
@@ -20,4 +20,4 @@
             "runtimeExecutable": "ts-node",
         }
     ]
-}
\ No newline at end of file
+}
```
GGonryun, line 23: nit, unnecessary change? I've noticed prettier will try to remove/add whitespace in json files. you can use the vs code command center to save without formatting to prevent these changes from happening.

src/util.ts
```diff
@@ -96,6 +97,15 @@ export const throwAssertNever = (value: never) => {
   throw assertNever(value);
 };
 
+// If the condition is true, aborts the process before throwing the error
+export const conditionalAbortBeforeThrow =
+  (abortBeforeThrow: boolean) => (err: any) => {
+    if (abortBeforeThrow) {
+      sys.exit(1);
```
GGonryun, line 104: nit: should we return after sys.exit to prevent the error from ever being thrown? 🤔  not sure if as is we can guarantee silent output.

src/commands/shared/request.ts
```diff
@@ -108,13 +108,19 @@ export const request =
...
-      fetchCommand<RequestResponse<T>>(resolvedAuthn, args, [
-        command,
-        ...args.arguments,
-      ])
-    );
+    const data =
+      options?.message != "quiet"
+        ? await spinUntil(
+            accessMessage(options?.message),
+            fetchCommand<RequestResponse<T>>(resolvedAuthn, args, [
```
GGonryun, line 115: DRY: it looks like the command we execute is the same so we can encapsulate the function as a variable and call it in the two places we need.

src/util.ts
bgberr: If we don't throw here, then we get a `| void` on the type of anything that uses this.  I would be very surprised if anything was running after sys.exit?

src/commands/ssh-resolve.ts
```diff
@@ -0,0 +1,119 @@
...
+        }),
+
+    fsShutdownGuard(sshResolveAction)
+  );
+
+/** Connect to an SSH backend
+ *
+ * Implicitly gains access to the SSH resource if required.
+ *
+ * Supported SSH mechanisms:
```
GGonryun, line 62: I don't think I fully understand this comment, i don't think resolve connects to anything. doesn't it just check to see if we can resolve a node's host name and make an access request transparently if we need to? And don't we also support gcloud and azure? 🤔

src/commands/ssh-resolve.ts
```diff
@@ -0,0 +1,119 @@
...
+ * - AWS EC2 via SSM with Okta SAML
+ */
+const sshResolveAction = async (
+  args: yargs.ArgumentsCamelCase<SshResolveCommandArgs>
+) => {
+  const p0Executable = bootstrapConfig.appPath;
+  // Prefix is required because the backend uses it to determine that this is an AWS request
```
GGonryun, line 69: this comment also doesn't make any sense to me, i don't see prefix being used here.

src/commands/ssh-resolve.ts
bgberr: Oops. Some comments left over in here from where I grabbed the boilerplate 🫣 Will clean up

src/commands/ssh-resolve.ts
```diff
@@ -0,0 +1,119 @@
...
+  const authn = await authenticate({ noRefresh: args.quiet ?? false }).catch(
+    conditionalAbortBeforeThrow(args.quiet ?? false)
```
GGonryun, line 71: nit: you can dry this by assigning the output of `conditionalAbortBeforeThrow(args.quiet ?? false)` to a variable.

```
const silently = conditionalAbortBeforeThrow(args.quiet ?? false)
await authenticate(...).catch(silently)
```

src/commands/ssh-resolve.ts
```diff
@@ -0,0 +1,119 @@
...
+    authn,
+    args,
+    args.destination,
+    true,
+    args.quiet
+  ).catch(conditionalAbortBeforeThrow(args.quiet ?? false));
+
+  const sshProvider = SSH_PROVIDERS[provisionedRequest.permission.provider];
+
+  const keys = await sshProvider?.generateKeys?.(
```
GGonryun, line 84: It doesn't look like generate keys actually creates anything? should we rename this to something that doesn't imply we're creating a resource?

src/commands/ssh-resolve.ts
```diff
@@ -0,0 +1,136 @@
...
+  print2("=".repeat(80));
+
+  await fs.promises.mkdir(path.join(P0_PATH, "ssh", "configs"), {
+    recursive: true,
+  });
+  const configLocation = path.join(
+    P0_PATH,
+    "ssh",
+    "configs",
+    `${args.destination}.config`
```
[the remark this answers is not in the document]
GGonryun: Agree with semgrep here, the destination could include characters that would cause the destination to resemble a path, we should clean them up before calling path.join.

src/commands/ssh-resolve.ts
bgberr: for Azure it will generate the identity & certificate files which is where the name came from, but there's no need to generate those for AWS/GCP

src/util.ts
GGonryun: i see, yeah i would be too tbh but wasn't sure if it's guaranteed. we can keep it like this for now.

src/plugins/okta/aws.ts
```diff
@@ -26,7 +26,7 @@ export const assumeRoleWithOktaSaml = async (
       );
       const { roles } = rolesFromSaml(account, samlResponse);
       if (!roles.includes(args.role))
-        throw `Role not available. Available roles:\n${roles.map((r) => `  ${r}`).join("\n")}`;
+        throw `Role ${args.role} not available. Available roles:\n${roles.map((r) => `  ${r}`).join("\n")}`;
```
GGonryun, line 29: thanks! 👍  should help with debugging.

src/plugins/aws/ssh.ts
```diff
@@ -113,6 +114,12 @@ export const awsSshProvider: SshProvider<
     return undefined;
   },
 
+  generateKeys: async (_) => {
+    return {
+      privateKeyPath: path.join(P0_PATH, "ssh", "id_rsa"),
```
GGonryun, line 119: DRY should we be using: `PRIVATE_KEY_PATH` here instead?  or should we be calling this method anywhere where we need to access the private key path for an ssh provider?

src/commands/ssh-resolve.ts
GGonryun: Ah, that makes sense, i saw that later in the PR lol.

src/plugins/aws/ssh.ts
```diff
@@ -87,19 +88,19 @@ export const awsSshProvider: SshProvider<
...
+  proxyCommand: (request, port) => {
     return [
       "aws",
       "ssm",
       "start-session",
       "--region",
       request.region,
       "--target",
-      "%h",
+      request.id,
```
GGonryun, line 99: why do you need to specify `request.id` here? wouldn't `%h` be enough since we add the hostname to the config?

src/plugins/azure/ssh.ts
```diff
@@ -92,7 +92,8 @@ export const azureSshProvider: SshProvider<
   },
 
   // Azure doesn't support ProxyCommand, as nice as that would be. Yet.
```
GGonryun, line 94: outdated comment.

src/commands/shared/ssh.ts
```diff
@@ -139,14 +155,17 @@ export const provisionRequest = async (
...
 
   if (!response) {
-    print2("Did not receive access ID from server");
+    if (!quiet) {
+      print2("Did not receive access ID from server");
+    }
     return;
   }
   const { id, isPreexisting } = response;
+  if (!quiet) print2("Did not receive access ID from server");
```
GGonryun, line 168: do we need this twice? 

```
miguelcampos@Miguels-MacBook-Pro p0cli % /Users/miguelcampos/src/p0cli/p0 ssh-resolve tiny-node
Did not receive access ID from server
Existing access found.  Connecting to instance.
```

src/commands/ssh-resolve.ts
```diff
@@ -0,0 +1,119 @@
...
+  );
+
+  fs.writeFileSync(configLocation, data);
```
GGonryun, line 118: we should add some debug logs for this command so that it gets easier to troubleshoot, i was having a hard time connecting and removing the `-q` flag didn't provide enough clarity about why my configs weren't matching.

src/plugins/ssh/index.ts
```diff
@@ -478,3 +479,49 @@ export const sshOrScp = async (args: {
...
   }
 };
+
+export const sshProxy = async (args: {
+  authn: Authn;
+  request: SshRequest;
+  cmdArgs: SshProxyCommandArgs;
+  privateKey: string;
+  sshProvider: SshProvider<any, any, any, any>;
+  debug: boolean;
```
GGonryun, line 489: how do i get a --debug flag passed down to my proxy commands? 🤔 this is my config, but since the generated configs are added on the fly i'm not sure how to debug with `ssh`, maybe we shouldn't need to 🤷 ? 

```
Match exec "/Users/miguelcampos/src/p0cli/p0 ssh-resolve %h -q"
 Include /Users/miguelcampos/.p0-development/ssh/configs/*.config
```

src/commands/ssh-resolve.ts
```diff
@@ -0,0 +1,119 @@
...
+  );
+
+  const { request, provisionedRequest } = await prepareRequest(
+    authn,
+    args,
+    args.destination,
+    true,
+    args.quiet
+  ).catch(conditionalAbortBeforeThrow(args.quiet ?? false));
+
```
GGonryun, line 81: nit: We should spit out the request as json after prepareRequest is called if the command isn't set to quiet, it'll help with debugging. 🤔

src/commands/ssh-proxy.ts
```diff
@@ -0,0 +1,101 @@
...
+import { authenticate } from "../drivers/auth";
+import { fsShutdownGuard } from "../drivers/firestore";
+import { sshProxy } from "../plugins/ssh";
+import { P0_PATH } from "../util";
+import { SSH_PROVIDERS, SshProxyCommandArgs } from "./shared/ssh";
+import * as fs from "fs/promises";
+import path from "path";
+import yargs from "yargs";
+
+export const sshProxyCommand = (yargs: yargs.Argv) =>
```
GGonryun, line 20: how would users request sudo access? 🤔

GGonryun: i just tested this and it looks like all you have to do is request sudo access for the node and when you ssh with the proxy command you'll have sudo too.

GGonryun: i guess the problem here is users can't ask for sudo when they ssh, they must ask for it beforehand. 🤔

src/plugins/ssh/index.ts
bgberr: If you run `shh-resolve` with `--debug` should we just forward that to `ssh-proxy`?

.vscode/launch.json
```diff
@@ -20,4 +20,4 @@
...
```
GGonryun, line 23: nit: might need to undo this change.

src/plugins/ssh/index.ts
GGonryun: maybe we can enhance this capability later, but i suppose for now we can debug with `p0 ssh`

GGonryun (approved):
changes are good 👍

PR #162, opened 2025-01-30
ENG-3314 Add support for per-terminal login

fabgo:
* Adds support for per-terminal logins
* Separate identity & config files are created for each terminal
* Each new terminal window requires a separate p0 login

### DIFF
--- a/src/commands/login.ts
+++ b/src/commands/login.ts
@@ -8,27 +8,20 @@ This file is part of @p0security/cli
 
 You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
 **/
-import {
-  authenticate,
-  IDENTITY_CACHE_PATH,
-  IDENTITY_FILE_PATH,
-} from "../drivers/auth";
+import { authenticate, writeIdentity } from "../drivers/auth";
 import { saveConfig } from "../drivers/config";
 import { fsShutdownGuard, initializeFirebase } from "../drivers/firestore";
 import { doc } from "../drivers/firestore";
 import { print2 } from "../drivers/stdio";
 import { pluginLoginMap } from "../plugins/login";
-import { TokenResponse } from "../types/oidc";
 import { OrgData, RawOrgData } from "../types/org";
 import { getDoc } from "firebase/firestore";
-import * as fs from "fs/promises";
-import * as path from "path";
 import yargs from "yargs";
 
 /** Logs in the user
  *
- * Currently only supports login to a single organization. Login credentials, together
- * with organization details, are saved to {@link IDENTITY_FILE_PATH}.
+ * Each terminal window requires a separate login. Login credentials,
+ * together with organization details, are saved to /tmp/p0.
  */
 export const login = async (
   args: { org: string },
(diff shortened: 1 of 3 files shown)


src/drivers/config.ts
```diff
@@ -41,17 +43,19 @@ export const saveConfig = async (orgId: string) => {
...
-  print2(`Saving config to ${CONFIG_FILE_PATH}.`);
+  const configFilePath = getConfigFilePath();
 
-  const dir = path.dirname(CONFIG_FILE_PATH);
+  print2(`Saving config to ${configFilePath}.`);
+
+  const dir = path.dirname(configFilePath);
   await fs.mkdir(dir, { recursive: true });
-  await fs.writeFile(CONFIG_FILE_PATH, JSON.stringify(config), { mode: "600" });
+  await fs.writeFile(configFilePath, JSON.stringify(config), { mode: "600" });
```
GGonryun, line 52: cc: @bgberr, I'm a little concerned with how this interacts with proxycommand 🤔. might be worth syncing up tomorrow if ya'll have a minute during collab.

src/drivers/config.ts
```diff
@@ -9,15 +9,17 @@ This file is part of @p0security/cli
...
 import { print2 } from "./stdio";
 import { getDoc } from "firebase/firestore";
 import fs from "fs/promises";
+import os from "os";
 import path from "path";
+import process from "process";
 
-export const CONFIG_FILE_PATH = path.join(P0_PATH, "config.json");
+const getConfigFilePath = () =>
+  path.join(os.tmpdir(), "p0", `config.json-${process.ppid}`);
```
GGonryun, line 22: is the parent id always going to be the terminal we have open, do we have to login again _every_ time we open a new terminal even if it's for the same environment?

GGonryun: jumped to the code before reading the description and yes we'll have to reauth every time we open a new terminal 🤔 . Idk if that will be that bad tho.

src/commands/login.ts
```diff
@@ -8,27 +8,20 @@ This file is part of @p0security/cli
...
 
 /** Logs in the user
  *
- * Currently only supports login to a single organization. Login credentials, together
- * with organization details, are saved to {@link IDENTITY_FILE_PATH}.
+ * Each terminal window requires a separate login. Login credentials,
+ * together with organization details, are saved to /tmp/p0.
  */
 export const login = async (
   args: { org: string },
```
GGonryun, line 27: have we considered using environment variables instead of a json config file? if we have to log in again anyways it might be easier to just save the env variables to the terminal session. 🤔

src/drivers/auth.ts
```diff
@@ -10,28 +10,32 @@ You should have received a copy of the GNU General Public License along with @p0
...
+import * as os from "os";
 import * as path from "path";
 
-export const IDENTITY_FILE_PATH = path.join(P0_PATH, "identity.json");
-export const IDENTITY_CACHE_PATH = path.join(
-  path.dirname(IDENTITY_FILE_PATH),
-  "cache"
-);
+const getIdentityFilePath = () =>
+  path.join(os.tmpdir(), "p0", `identity-${process.ppid}.json`);
```
GGonryun, line 22: nit: maybe we can make the path `{temp_dir}/p0/{ppid}/identity.json`?

src/commands/login.ts
fabgo: I researched this a bit, and it's tricky to do, since we have to set the environment variables on the parent (terminal) process.

PR #169, opened 2025-02-13
Remove public key requirement for SSH sessions access

GGonryun:
This PR moves the public key requirement for SSH session access from the request time to access time. At access time a user calls the `integrations/ssh/public-keys` endpoint which, for aws only, will let us upload SSH keys to the target machine. The exercise grant endpoint validates that the user has the correct permissions to access the node before placing their SSH key on the machine.


src/commands/shared/ssh.ts
```diff
@@ -194,17 +195,10 @@ export const prepareRequest = async (
     throw "Server did not return a request id. Please contact support@p0.dev for assistance.";
   }
 
-  const { provisionedRequest, publicKey } = result;
+  const { provisionedRequest } = result;
 
   const sshProvider = SSH_PROVIDERS[provisionedRequest.permission.provider];
 
-  if (
-    sshProvider.validateSshKey &&
```
GGonryun, line 202: we can no longer guarantee that a permission will include a valid ssh key because it can be uploaded at access time.

src/plugins/aws/ssh.ts
```diff
@@ -88,6 +86,16 @@ export const awsSshProvider: SshProvider<
 
   preTestAccessPropagationArgs: () => undefined,
 
+  async submitPublicKey(authn, request, requestId, publicKey) {
+    if (request.generated.publicKey) {
+      if (request.generated.publicKey !== publicKey) {
+        throw "Public key mismatch. Please revoke the request and try again.";
```
fabgo, line 92: What does it mean to revoke a request?

fabgo (approved):

PR #172, opened 2025-03-11
ENG-3667 Support google-oidc login

fabgo:
If the SSO provider is "google-oidc", the p0 login command currently fails. This PR ensures that the login flow works correctly, and that accounts using "google-oidc" as the login provider can login successfully.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@p0security/cli",
-  "version": "0.14.0",
+  "version": "0.14.1",
   "description": "Execute infra CLI commands with P0 grants",
   "main": "index.ts",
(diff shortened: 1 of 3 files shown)


GGonryun (approved):

gergas3 (approved):
We can merge this. I think we also need another Linear ticket to fix this in the backend. Here in the CLI we already have a handler for the `oidc-pkce` ssoProvider, but it seems we are not setting that up for users the sign in with Google login.
If we migrated the `google-oidc` (and `azure-oidc`) ssoProviders to just `oidc-pkce` with providerType=google|azure then I think the CLI would just work.

PR #173, opened 2025-03-26
Add email to google openid scope

MichaelDimitras:
There is currently a bug that creates a user with a blank email in firestore when a user logs in for the first time.

This blank user breaks all subsequent logins.

I found that adding the email scope to the request for the OIDC token to google fixes this issue.  With this change logging in with the CLI creates a user with my email.

https://developers.google.com/identity/openid-connect/openid-connect#authenticationuriparameters
>If the email scope value is present, the ID token includes email and email_verified claims.


GGonryun (approved):

gergas3 (approved):
👏

fabgo:
Nice work on this @MichaelDimitras !

PR #175, opened 2025-04-09
p0 ls Machine readable output

bgberr:
Enables --json object to directly output json response from backend

### DIFF
--- a/src/commands/ls.ts
+++ b/src/commands/ls.ts
@@ -40,10 +40,15 @@ const lsArgs = <T>(yargs: yargs.Argv<T>) =>
       array: true,
       string: true,
       default: [] as string[],
+    })
+    .option("json", {
+      type: "boolean",
+      default: false,
+      description: "Output in JSON format",
     });
 
 export const lsCommand = (yargs: yargs.Argv) =>
-  yargs.command<{ arguments: string[] }>(
+  yargs.command<{ arguments: string[]; json: boolean }>(
     "ls [arguments..]",
     "List request-command arguments",
     lsArgs,
@@ -72,6 +77,7 @@ const convertLsSizeArg = (args: string[]) => {
 const ls = async (
   args: yargs.ArgumentsCamelCase<{
     arguments: string[];
+    json: boolean;
   }>
 ) => {
   const authn = await authenticate();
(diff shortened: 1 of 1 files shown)


src/commands/ls.ts
```diff
@@ -72,18 +77,27 @@ const convertLsSizeArg = (args: string[]) => {
...
+    json: boolean;
   }>
 ) => {
   const authn = await authenticate();
   const { convertedArgs, requestedSize } = convertLsSizeArg(args.arguments);
 
   const data = await spinUntil(
     "Listing accessible resources",
-    fetchCommand<LsResponse>(authn, args, ["ls", ...convertedArgs])
+    fetchCommand<LsResponse>(authn, args, [
```
GGonryun, line 88: im not fully understanding why we need to send the json parameter to the backend? it doesn't seem to change anything there does it? 🤔

src/commands/ls.ts
```diff
@@ -72,18 +77,27 @@ const convertLsSizeArg = (args: string[]) => {
...
+      "ls",
+      ...(args.json ? args.arguments : convertedArgs),
+    ])
   );
-  const allArguments = [...args._, ...args.arguments];
-
   if (data && "ok" in data && data.ok) {
+    if (args.json) {
```
GGonryun, line 94: what happens if we have a really big list of like 1000 nodes? do we truncate the list in the response? do we have to support pagination of some sort?

bgberr: The pagination is the same as whatever it is for the existing p0 ls, which I think is to just show 30 nodes and to indicate it's truncated.  Will try to implement something better

src/commands/ls.ts
bgberr: Just didn't see any reason to strip it out.  The backend doesn't currently use it but it includes it for the sake of having it appear in the help output

src/commands/ls.ts
```diff
@@ -72,18 +77,27 @@ const convertLsSizeArg = (args: string[]) => {
...
+      print2(JSON.stringify(data, null, 2));
+      return;
+    }
+
+    const allArguments = [...args._, ...args.arguments];
+
     const label = pluralize(data.arg);
```
fabgo, line 101: Nit: Is label used outside line 103? Maybe move this inside the if statement? I know it's not related to your PR.

fabgo (approved):

src/commands/ls.ts
bgberr: Used on 89 as well

GGonryun (approved):

PR #177, opened 2025-04-15
ENG-4099 Support `p0 aws assume permission-set` command

fabgo:
* Added a `p0 aws assume permission-set` command for AWS IDC.
* The `p0 aws assume` command will now request access before assuming the role / permission-set as needed.
* Removed the `p0 aws role ls` command. Use the `p0 ls aws role name` command instead. (Since `p0 aws role assume` now requests access, if a role is marked with * when executing `p0 ls aws role name`, access will be automatically granted if not already.)


nbrahms:
I don't understand why this is removing `p0 aws role ls` as part of this change.

This is so the user can list the roles that are currently assumable, which is fundamentally different than what `p0 ls aws role name` does.

nbrahms:
For the purposes of review, could you please separate the refactor part of this work fro the functional change? AFAICT a fair amount of this is moved code, but it's difficult to separate out what is moved vs. what is new / changed.

src/plugins/aws/idc/index.ts
```diff
@@ -187,6 +187,7 @@ export const assumeRoleWithIdc = async (args: {
         accountId: args.accountId,
         permissionSet: args.permissionSet,
       });
+
```
GGonryun, line 190: 💭

src/plugins/aws/config.ts
```diff
@@ -14,6 +14,22 @@ import { AwsConfig } from "./types";
 import { getDoc } from "firebase/firestore";
 import { sortBy } from "lodash";
 
+export const getFirstAwsConfig = async (authn: Authn) => {
```
GGonryun, line 17: i feel like this should throw an error if we're installed on multiple accounts. i'm worried people may get confused as to why certain behaviors work because we're just assuming a config without using an explicit account with `--account`

src/index.ts
```diff
@@ -9,13 +9,19 @@ This file is part of @p0security/cli
...
 import { noop } from "lodash";
 
-export const main = () => {
-  // We can suppress output here, as .fail() already print2 errors
-  void (cli.parse() as any).catch(noop);
+export const main = async () => {
+  await initializeFirebase();
+
+  try {
+    await cli.parseAsync();
```
GGonryun, line 19: nit: i think you can still do`await cli.parseAsync().catch(noop)` instead of adding a try catch here.

src/commands/aws/util.ts
```diff
@@ -0,0 +1,54 @@
...
+import { request } from "../shared/request";
+import yargs from "yargs";
+
+export const PROVISIONING_ACCESS_MESSAGE =
+  "Waiting for access to be provisioned";
+export const EXISTING_ACCESS_MESSAGE = "Existing access found.";
+export const ACCESS_EXISTS_ERROR_MESSAGE =
+  "This principal already has this access";
+
+export const provisionRequest = async (
```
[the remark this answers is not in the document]
GGonryun: also can some of this be DRY'd with `provisionRequest` in commands/ssh 🤔

src/commands/aws/index.ts
```diff
@@ -8,19 +8,34 @@ This file is part of @p0security/cli
...
-const awsCommands = [role];
-
-const awsArgs = (yargs: yargs.Argv) => {
+const awsArgs = async (yargs: yargs.Argv) => {
   const base = yargs
     .option("account", {
       type: "string",
       describe: "AWS account ID or alias (or set P0_AWS_ACCOUNT)",
     })
+    .option("reason", {
```
GGonryun, line 23: what if i already have access to something? Do i still need to pass in a reason every time?

GGonryun (approved):
lgtm, some questions for my understanding and nits. no strong opinions on them tho.

src/commands/aws/index.ts
fabgo: It's an optional argument.

src/plugins/aws/config.ts
fabgo: This is to grab the authentication method, e.g. IAM vs. IDC. It is assumed to be the same across all accounts.

Re: our previous discussion regarding this.

src/index.ts
fabgo: Done.

src/commands/aws/util.ts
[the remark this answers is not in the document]
fabgo: Implementation is a little different. But I can move it.

PR #185, opened 2025-05-06
ENG-2713 Make the contact/help message configurable

fabgo:
Provide a configurable contact/help message so that we can provide a custom message to tenants if needed.

### DIFF
--- a/src/commands/index.ts
+++ b/src/commands/index.ts
@@ -8,7 +8,8 @@ This file is part of @p0security/cli
 
 You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
 **/
-import { print2 } from "../drivers/stdio";
+import { getHelpMessage } from "../drivers/config";
+import { print1, print2 } from "../drivers/stdio";
 import { checkVersion } from "../middlewares/version";
 import { allowCommand } from "./allow";
 import { awsCommand } from "./aws";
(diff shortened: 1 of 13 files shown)


src/commands/index.ts
```diff
@@ -41,8 +42,23 @@ const commands = [
   kubeconfigCommand,
 ];
 
+const argv = yargs(hideBin(process.argv));
+
+// Override the default yargs showHelp function to include a custom help message at the end
+const originalShowHelp = argv.showHelp.bind(argv);
```
GGonryun, line 48: nit: it'd be nice to encapsulate this all in a function 🤔

src/commands/ssh-resolve.ts
```diff
@@ -124,7 +124,7 @@ const sshResolveAction = async (
     ? `CertificateFile ${keys.certificatePath}`
     : "";
 
-  const p0Executable = bootstrapConfig.appPath;
+  const p0Executable = env.P0_APP_PATH ?? "p0";
```
GGonryun, line 127: nit: by any chance do we have a `getAppPath()` method? this feels like we might need it in other places

src/index.ts
```diff
@@ -9,9 +9,21 @@ This file is part of @p0security/cli
 You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
 **/
 import { cli } from "./commands";
+import { loadConfig } from "./drivers/config";
 import { noop } from "lodash";
 
 export const main = async () => {
+  // Try to load the config early here to get the custom help/contact messages (if any)
+  try {
+    await loadConfig();
```
GGonryun, line 18: can you explain why this is necessary? don't we download the latest config after they log in for the first time?

GGonryun (approved):

src/index.ts
fabgo: OK, will make it more clear.

src/commands/index.ts
fabgo: Done

src/commands/ssh-resolve.ts
fabgo: Done

PR #186, opened 2025-05-06
Single-artifact CLI builds

dhing1024:
**Problem**

Some customers are interested in a CLI build that can be installed without having to install NodeJS or any of our other dependencies independently, to reduce the supply-chain attack surface area.

**Change**

Adds some scripts to generate a CLI artifact using NodeJS's experimental single executable application functionality. See https://nodejs.org/docs/latest-v20.x/api/single-executable-applications.html

**Validation**

To build:
- Run `yarn build` to compile typescript
- From the root directory, run `source build.sh` to generate the CLI artifact

To test binary:
- Run `nvm deactivate` to temporarily remove NodeJS from your path
- Ensure that `which node` indicates that node cannot be found
- Run the p0 artifact to login and make a request

To test install via NPM:
- Remove global npm install
- Run `yarn build` && `npm pack`
- Run `npm install -g <npm_packed_file>`
- Run the p0 command to login and make a request

**Tracking**

https://linear.app/p0-security/issue/ENG-4341/package-p0-cli-into-a-single-deployable-artifact

**Implementation**

- Moves all build artifacts, including the tsc output, to a `/build` directory.
- The `source build.sh` command is outside of the existing `yarn build` script, because for some reason it wouldn't work if I included in there.
- The build is currently only tested on MacOS. Some of the build commands might need to be modified for Windows and Linux (see the NodeJS documentation for details)
- As-is, the CLI artifact generated by these scripts may not work if you move it to a different device, due to code signing.


p0.js
GGonryun, line 1: why do we need this `p0.js` instead of the old p0 artifact? 🤔

dhing1024: If I remember correctly, one of the complications is that the file in the `main` field of `sea-config.json` can only `require()` built-in NodeJS modules. You need to use `createRequire()` to import anything else.

nbrahms:
👋  Have you confirmed that running `p0` after installing from`npm` would still work?

Naively I'd expect that it wouldn't as this seems to delete the script.

dhing1024:
> 👋 Have you confirmed that running `p0` after installing from`npm` would still work?
> 
> Naively I'd expect that it wouldn't as this seems to delete the script.

I originally thought it might be complicated to support both the single-deployable artifact and install from NPM, but I think it might actually be easier than I thought. I will add it back and test that

build.sh
```diff
@@ -0,0 +1,11 @@
+cp ./p0.js ./build/p0.js
+
+# See https://nodejs.org/docs/latest-v20.x/api/single-executable-applications.html for more information.
+# Some of these commands only work on macOS. Windows and Linux require slightly different commands.
```
GGonryun, line 4: Does this mean that the executable we build is only runnable on mac machines with our same chip: `Apple M3`? 🤔

.eslintrc.js
```diff
@@ -62,6 +62,7 @@ module.exports = {
     "dist/**",
```
GGonryun, line 62: nit: should we ignore `build/**` now 🤔

build.sh
dhing1024: The way it works is instead of building an executable from scratch, it uses whatever version of NodeJS is already installed on your system and putting your custom application code (in this case our CLI) into a copy of the existing node binary. So my guess is that the built P0 CLI would be runnable on machines that the underlying Node binary is compatible with.

.eslintrc.js
dhing1024: Yes

build.sh
GGonryun: I wonder if this means we'll need to start trying to build and host deployable's somewhere 🤔.

I don't think we need to worry about distribution at this time but something to keep in mind going forward. Like if we publish an update we'll need CI/CD to create a deployable, save it somewhere, and we'll need to make sure the version check works with the packaged executable and tells the user correctly where to download the latest version from.

GGonryun (approved):

GGonryun:
It would also be useful to include instructions on how to build the CLI as a single executable in the contributing.md doc and/or a separate package.json command to build for mac's on our chipset.

gergas3 (approved):

gergas3 (commented):
I had to make the `p0` and `build.sh` files executable when testing (chmod +x). That seems can be committed to git.
```
build % git diff
diff --git a/build.sh b/build.sh
old mode 100644
new mode 100755
diff --git a/p0 b/p0
old mode 100644
new mode 100755
```

esbuild.js
```diff
@@ -0,0 +1,20 @@
...
+
+const esbuild = require("esbuild");
+
+esbuild
+  .build({
+    entryPoints: ["./build/dist/index.js"],
+    bundle: true,
+    platform: "node",
+    target: "node20",
+    format: "cjs",
```
dhing1024, line 10: The output format needs to be CommonJS because Node-SEA doesn't support ESM. See https://nodejs.org/api/single-executable-applications.html#single-executable-applications

sea-config.json
```diff
@@ -1,5 +1,9 @@
 {
   "main": "./build/p0.js",
   "output": "./build/p0.blob",
-  "disableExperimentalSEAWarning": true
+  "disableExperimentalSEAWarning": true,
+  "assets": {
+    "favicon.ico": "./public/favicon.ico",
+    "redirect-landing.html": "./public/redirect-landing.html"
```
dhing1024, line 7: Unfortunately I don't think there's a way to do this without having to explicitly enumerate the static assets we want to include individually. These are used as the landing webpage post-login.

package.json
```diff
@@ -49,7 +49,7 @@
     "@types/jest": "^29.5.12",
     "@types/jsdom": "^21.1.6",
     "@types/lodash": "^4.14.202",
-    "@types/node": "^18.11.7",
+    "@types/node": "^20.17.46",
```
dhing1024, line 52: Bumping this for correct type-annotations from `node:sea`

src/common/auth/server.ts
```diff
@@ -12,24 +12,49 @@ You should have received a copy of the GNU General Public License along with @p0
...
+};
+
 /** Waits for an OIDC authorization redirect using a locally mounted server */
 export const withRedirectServer = async <S, T, U>(
   start: (server: http.Server) => Promise<S>,
   complete: (value: S, token: T) => Promise<U>,
   options?: { port?: number }
 ) => {
   const app = express();
-  app.use(express.static(`${ROOT_PATH}/public`));
```
dhing1024, line 32: Express.static assets aren't bundled into the executable, so a different solution is needed here. Otherwise the landing page after login won't work.

esbuild.js
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env node
+
+const esbuild = require("esbuild");
+
+esbuild
+  .build({
+    entryPoints: ["./build/dist/index.js"],
+    bundle: true,
```
dhing1024, line 7: Bundling is the primary reason for adding esbuild, as Node-SEA only allows for a single script to be included in the binary. See https://nodejs.org/api/single-executable-applications.html#single-executable-applications

This means we need an extra step to convert everything into one file, and use that file as the entrypoint to the final binary.

dhing1024 (commented):
I think this fixes some bugs related to this feature. There is one other minor thing: when I run the application I get this warning:
```
Warning: Currently the require() provided to the main script embedded into single-executable applications only supports loading built-in modules.
To load a module from disk after the single executable application is launched, use require("module").createRequire().
Support for bundled module loading or virtual file systems are under discussions in https://github.com/nodejs/single-executable
(Use `p0 --trace-warnings ...` to show where the warning was created)
```
The "main script" it refers to is the output `p0.js` file created by esbuild, and I'm not sure (1) if this is a benign issue or if it has some effect on us, and (2) if it is benign, how to validate that. If it isn't a problem, I was thinking I'd just monkey-patch to silence the warning.

esbuild.js
fabgo: This should probably be a code comment.

esbuild.js
fabgo: This also would be useful to have as a code comment.

GGonryun:
> Warning: Currently the require() provided to the main script embedded into single-executable applications only supports loading built-in modules.
To load a module from disk after the single executable application is launched, use require("module").createRequire().
Support for bundled module loading or virtual file systems are under discussions in https://github.com/nodejs/single-executable
(Use `p0 --trace-warnings ...` to show where the warning was created)

Do you get anything when you run the recommended command:

```
p0 --trace-warnings 
```

I don't think we want to show this warning to the customer, it comes up for me every single time i run any p0 command.

dhing1024:
> Do you get anything when you run the recommended command:
> 
> ```
> p0 --trace-warnings 
> ```
> 
> I don't think we want to show this warning to the customer, it comes up for me every single time i run any p0 command.

If you use `p0 --trace-warnings` that'll just bring up our help message from the CLI and doesn't do anything else. I did try `NODE_OPTIONS="--trace-warnings" p0 ...` but it prints the same warning message _except_ for the `(Use p0 --trace-warnings ... to show where the warning was created)` part.

CONTRIBUTING.md
```diff
@@ -21,3 +21,10 @@ follow certain guidelines, in order to ensure usability of the CLI:
     via `console.log`
 - Client-usage errors (that is, errors that provide feedback to the user of expected CLI misuse)
   should be emmitted using `throw <message>`; this prevents a stack trace from being shown
+
+## Single-Artifact Binary
+
+To build the CLI as a single-artifact binary, use the `yarn build:sea` script and then
+run `build.sh`. You must have NodeJS version 22 or newer currently installed on your system
```
fabgo, line 28: Don't forget to update README.md as well.

src/common/auth/server.ts
```diff
@@ -39,12 +64,22 @@ export const withRedirectServer = async <S, T, U>(
...
+  });
+
+  // handle redirect
   const redirectRouter = express.Router();
   redirectRouter.get("/", (req, res) => {
     const token = req.query as T;
     complete(value, token)
       .then((result) => {
-        res.status(200).sendFile(`${ROOT_PATH}/public/redirect-landing.html`);
+        pipeToResponse(pageBytes, res, "text/html; charset=utf-8");
```
fabgo, line 82: Do we not set the response status code any more?

src/plugins/okta/login.ts
```diff
@@ -71,11 +71,9 @@ const fetchSamlResponse = async (
...
   await validateResponse(response);
   const html = await response.text();
-  const dom = new JSDOM(html);
-  const samlInput = dom.window.document.querySelector(
-    'input[name="SAMLResponse"]'
-  );
-  return (samlInput as HTMLInputElement | undefined)?.value;
+  const $ = cheerio.load(html);
+  const samlInputValue = $('input[name="SAMLResponse"]').val();
+  return typeof samlInputValue === "string" ? samlInputValue : undefined;
```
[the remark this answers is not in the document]
fabgo: Was this resolved? This would be important to test before merging.

fabgo (commented):
Looks good, but verifying that this works on Okta before merging would be prudent.

src/common/auth/server.ts
dhing1024: Express automatically sets a status 200 if not explicitly set, but I can add it back for completeness

src/plugins/okta/login.ts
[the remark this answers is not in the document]
dhing1024: Tested this out and it works

PR #187, opened 2025-05-09
ENG-4641 Catch and handle authentication errors

fabgo:
Catch and handle authentication errors:

* Inform end-user of stale/invalid token errors
* Provide a public help message for unexpected auth errors

### DIFF
--- a/src/drivers/auth/index.ts
+++ b/src/drivers/auth/index.ts
@@ -14,6 +14,7 @@ import { TokenResponse } from "../../types/oidc";
 import { OrgData } from "../../types/org";
 import { authenticateToFirebase } from "../firestore";
 import { print2 } from "../stdio";
+import { EXPIRED_CREDENTIALS_MESSAGE } from "../util";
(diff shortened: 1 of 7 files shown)


GGonryun (approved):

src/drivers/auth/index.ts
```diff
@@ -108,6 +98,26 @@ const loadCredentialsWithAutoLogin = async (options?: {
...
 };
 
+export const remainingTokenTime = (identity: Identity) =>
+  identity.credential.expires_at - Date.now() * 1e-3;
+
+const loadCredentialsWithAutoLogin = async (options?: {
+  noRefresh?: boolean;
+}): Promise<Identity> => {
+  const identity = await loadCredentials();
+  if (remainingTokenTime(identity) > 60) {
```
GGonryun, line 108: tiniest of nits, but we could save this magic value somewhere so that it's easier to handle in the future and possible also self-document like if these are MS or S.

PR #189, opened 2025-05-13
Bump required Node engine version to 22

dhing1024:
The updated CLI that supports the deployable artifact requires a minimum of 20.12 (on major version 20) or 20.7 (on major version 21). This is not an issue on Node 22+. To accommodate these new requirements, we will bump our required node engine to 22.

While we could support Node 21 for deploying with NPM and use Node 22 for the standalone CLI, we believe it may be cleaner to simply bump the version to 22. Building the standalone CLI with Node 20 also results in some build warnings, so there may also be some stability benefits.


fabgo (approved):

package.json
```diff
@@ -13,7 +13,7 @@
     "p0": "p0"
   },
   "engines": {
-    "node": ">=20.14"
+    "node": ">=22"
```
GGonryun, line 16: do any of the docs need to change to reflect this update?

GGonryun (approved):

package.json
dhing1024: I updated the docs in #186

PR #190, opened 2025-05-14
Build standalone CLI executable in CI/CD on release creation

dhing1024:
Triggers a standalone binary executable build in CI/CD when releases get created. The generated binaries are uploaded to the corresponding release that triggered it. The CI/CD build includes the build process for the executable, codesigning for MacOS, and notarization by Apple. Only MacOS is currently supported.


.github/workflows/build-standalone.yml
```diff
@@ -0,0 +1,74 @@
...
+jobs:
+  build_cli:
+    name: Build CLI on MacOS
+    runs-on: macos-latest
+    env:
+      APPLE_APP_STORE_CONNECT_KEY_ID: ${{ secrets.APPLE_APP_STORE_CONNECT_KEY_ID }}
+      APPLE_APP_STORE_CONNECT_ISSUER_ID: ${{ secrets.APPLE_APP_STORE_CONNECT_ISSUER_ID }}
+      APPLE_CODESIGN_CERT_ID: "F9F7EEE981D7A60F83EBEC36AD675448CA0AB41B"
+    steps:
+      - uses: actions/checkout@v3
```
aceat64, line 16: Please use `actions/checkout@v4` instead.

.github/workflows/build-standalone.yml
```diff
@@ -0,0 +1,74 @@
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ^22
+          cache: yarn
+          cache-dependency-path: |
+            yarn.lock
```
aceat64, line 22: Use `actions/setup-node@v4`, it would also be best to use `node-version-file: '.nvmrc'` instead of `node-version: ^22`. You'll need to create the `.nvmrc` file, this can be done with `nvm use [<version>] --save`.

.github/workflows/build-standalone.yml
```diff
@@ -0,0 +1,74 @@
...
+name: Build Standalone CLI
+
+jobs:
+  build_cli:
+    name: Build CLI on MacOS
+    runs-on: macos-latest
+    env:
+      APPLE_APP_STORE_CONNECT_KEY_ID: ${{ secrets.APPLE_APP_STORE_CONNECT_KEY_ID }}
+      APPLE_APP_STORE_CONNECT_ISSUER_ID: ${{ secrets.APPLE_APP_STORE_CONNECT_ISSUER_ID }}
+      APPLE_CODESIGN_CERT_ID: "F9F7EEE981D7A60F83EBEC36AD675448CA0AB41B"
```
aceat64, line 14: Make this a repo variable so it can be used for multiple workflows.

Repo vars are similar to repo secrets, just not secret 😆 and can be referenced like this:
```
${{ vars. APPLE_CODESIGN_CERT_ID }}
```

If the other two IDs aren't secrets, they should be moved to repo vars as well.

.github/workflows/build-standalone.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: Get yarn cache directory path
+        id: yarn-cache-dir-path
+        run: |
+          echo yarn_dir="$(yarn cache dir)" >> $GITHUB_OUTPUT
+      - uses: actions/cache@v3
+        with:
+          path: |
+            **/node_modules
+            ${{ steps.yarn-cache-dir-path.outputs.yarn_dir }}
+          key: ${{ runner.os }}-yarn-frontend-${{ steps.yarn-cache-dir-path.outputs.yarn_dir }}-${{ hashFiles('**/yarn.lock') }}
```
aceat64, line 48: The `setup-node@v4` action will already be taking care of yarn's cache, so you can delete the `yarn-cache-dir-path` step and use this for the caching step:

```
      - uses: actions/cache@v4
        with:
          path: |
            ./node_modules
          key: ${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}
```

aceat64 (changes requested):

.github/workflows/build-standalone.yml
dhing1024: @aceat64 to make a follow-up ticket for this work at a later time

PR #192, opened 2025-05-14
Password login

gergas3:
Add support for email / password login to P0 backend.

The email and password are provided by the user via the `P0_EMAIL` and `P0_PASSWORD` environment variables to avoid storing these in the cached identity file.

If an org uses password login then their "login" is the same as authenticating to firebase. That, however is not deduplicated. `loadCredentialsWithAutoLogin` and `authenticateToFirebase` both sign in the user with email / password to firebase. Although this is redundant, keeping that logic makes the change much smaller and easier to see that existing functionality is not affected.

Manually tested the following cases:
- The P0_ORG environment variable continues to work
- Google login to P0 + AWS SSH access using AWS Identity Center with Okta
- email / password login + AWS SSH access using AWS Identity Center with Okta
- email / password login + AWS SSH access using AWS Account Federation with Okta:
    this errors out with
    ```{"error":"invalid_request","error_description":"'subject_token' is invalid."}```
    because we cannot sign in to the Okta AWS Account Federation app with a token from Firebase

Example commands:
Screenshot 2025-05-13 at 10 30 40 PM

Added the `--debug` flag to the `p0 login` command as well. This prints various Firebase errors, for instance, if the user has not signed up yet:

image

### DIFF
--- a/src/commands/login.ts
+++ b/src/commands/login.ts
@@ -27,8 +27,11 @@ import yargs from "yargs";
 const MIN_REMAINING_TOKEN_TIME_SECONDS = 5 * 60;
 
 const doActualLogin = async (orgWithSlug: OrgData) => {
-  const plugin = orgWithSlug?.ssoProvider;
-  const loginFn = pluginLoginMap[plugin];
+  const plugin =
+    orgWithSlug?.ssoProvider ??
+    (orgWithSlug.usePassword ? "password" : undefined);
+
+  const loginFn = plugin && pluginLoginMap[plugin];
 
   if (!loginFn) throw "Unsupported login for your organization";
 
@@ -53,8 +56,8 @@ const formatTimeLeft = (seconds: number) => {
  * Otherwise, the identity file is written to the ~/.p0 directory.
  */
 export const login = async (
-  args: { org?: string; refresh?: boolean },
-  options?: { skipAuthenticate?: boolean }
+  args: { org: string; refresh?: boolean },
+  options?: { debug?: boolean; skipAuthenticate?: boolean }
 ) => {
   let identity;
   try {
@@ -112,7 +115,7 @@ export const login = async (
   }
 
   if (!options?.skipAuthenticate) {
-    await authenticate();
+    await authenticate({ debug: options?.debug });
     await validateTenantAccess(orgData);
   }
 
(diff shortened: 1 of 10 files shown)


src/drivers/firestore.ts
```diff
@@ -45,35 +49,33 @@ export async function initializeFirebase() {
   }
 }
 
-const findProviderId = (identity: Identity) => {
-  switch (identity.org.ssoProvider) {
+const findProviderId = (org: OrgData) => {
+  switch (org.ssoProvider) {
     case "google":
-      return SignInMethod.GOOGLE;
+      return ProviderId.GOOGLE;
```
gergas3, line 55: Changed this to a providerId constant. The value of the constant is unchanged: "google.com".

src/types/oidc.ts
```diff
@@ -34,11 +34,11 @@ export type AuthorizeResponse = {
...
   id_token: string;
-  token_type: string;
-  scope: string;
+  token_type?: string;
+  scope?: string;
   expires_in: number;
-  refresh_token: string;
-  device_secret: string;
+  refresh_token?: string;
+  device_secret?: string;
```
gergas3, line 41: Made these optional because they currently aren't read anywhere, and Firebase auth doesn't return them.

src/commands/login.ts
```diff
@@ -53,8 +56,8 @@ const formatTimeLeft = (seconds: number) => {
  * Otherwise, the identity file is written to the ~/.p0 directory.
  */
 export const login = async (
-  args: { org?: string; refresh?: boolean },
-  options?: { skipAuthenticate?: boolean }
+  args: { org: string; refresh?: boolean },
```
GGonryun, line 59: I think this is intentionally left optional, users can provide the P0_ORG var or if left blank it will respond with whether or not the user is already logged in 🤔

src/commands/login.ts
```diff
@@ -137,8 +140,21 @@ export const loginCommand = (yargs: yargs.Argv) =>
...
-    fsShutdownGuard(login)
+    fsShutdownGuard(
+      (
+        args: yargs.ArgumentsCamelCase<{
+          org: string;
+          refresh?: boolean;
+          debug?: boolean;
+        }>
+      ) =>
+        login({ org: args.org, refresh: args.refresh }, { debug: args.debug })
```
GGonryun, line 156: nit: could we pass in login(args, args)?

GGonryun (approved):

src/commands/login.ts
gergas3: Thanks for finding that. Updated.

src/commands/login.ts
gergas3: 🤔 We could... it's shorter for sure.

PR #193, opened 2025-05-19
Print only AWS credentials on `p0 aws permission-set assume` command

fabgo:
The printAwsCredential() function was iterating over any field in the AwsCredential object, which include an extra `expiresAt` field used for only for caching.
    
Instead, changed it to print only the specific field values that contain the credentials.

### DIFF
--- a/src/commands/aws/__tests__/__snapshots__/role.test.ts.snap
+++ b/src/commands/aws/__tests__/__snapshots__/role.test.ts.snap
@@ -9,9 +9,8 @@ exports[`aws role a single installed account with Okta SAML assume should assume
   [
     "
 Or, populate these environment variables using BASH command substitution:
-
-  $(p0 aws role assume Role1)
-",
+  
+  $(p0 aws role assume Role1) ",
   ],
 ]
 `;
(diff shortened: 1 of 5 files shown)


src/commands/aws/__tests__/__snapshots__/role.test.ts.snap
```diff
@@ -9,20 +9,25 @@ exports[`aws role a single installed account with Okta SAML assume should assume
...
 `;
 
 exports[`aws role a single installed account with Okta SAML assume should assume a role: stdout 1`] = `
 [
   [
-    "  export AWS_ACCESS_KEY_ID=test-access-key
-  export AWS_SECRET_ACCESS_KEY=secret-access-key
-  export AWS_SESSION_TOKEN=session-token
-  export AWS_SECURITY_TOKEN=session-token",
+    "  export AWS_ACCESS_KEY_ID=test-access-key",
```
GGonryun, line 21: Is this intentional? It looks like it returns a list of arrays now instead of just a string with new lines?

gergas3 (approved):

src/commands/aws/__tests__/__snapshots__/role.test.ts.snap
fabgo: It uses a separate print1 statement for each now.

PR #194, opened 2025-05-19
ENG-4817 Add refresh token support

fabgo:
* Adds support for refresh tokens.
* Okta and Google OIDC are currently supported.
* Verified that token refresh works on both Google and Okta.
* Added unit tests to verify and  guard against regressionsm.


package.json
```diff
@@ -31,6 +31,7 @@
     "firebase": "^10.7.2",
     "ini": "^4.1.3",
     "inquirer": "^9.2.15",
+    "jsdom": "^26.1.0",
```
dhing1024, line 34: Is there a reason why you need `jsdom` to be added back here? I took it out in https://github.com/p0-security/p0cli/pull/186 in favor of `cheerio` because it was causing a build warning for the standalone CLI.

If you need some simple HTML parsing or basic DOM manipulation, `cheerio` should be sufficient. "jsdom" is a little overkill IMO.

src/plugins/ping/login.ts
```diff
@@ -10,18 +10,30 @@ You should have received a copy of the GNU General Public License along with @p0
 **/
 import { AuthorizeResponse, TokenResponse } from "../../types/oidc";
 import { OrgData } from "../../types/org";
+import { LoginPlugin, LoginPluginMethods } from "../login";
 import { oidcLogin, oidcLoginSteps } from "../oidc/login";
 
+const SCOPES = "openid email profile";
```
dhing1024, line 16: Does this need the `offline_access` scope as well?

PR #206, opened 2025-06-17
General improvements to CLI version management

dhing1024:
Running `p0 --version` using the standalone CLI prints "unknown". This is likely due to yargs' default --version parser not being able to locate package.json. That + a few other fixes.

- Makes it so the version in package.json is the source of truth (no need to change it in multiple places)
- Add a function that identifies the version in all environments
- Use this implementation everywhere we need the version number
- Fix a potential bug where the auto-update message may not display
- Improve docs and messaging indicating how to install / upgrade

Tested by building locally and seeing that the version prints itself and the CLI still runs

### DIFF
--- a/src/commands/index.ts
+++ b/src/commands/index.ts
@@ -11,6 +11,7 @@ You should have received a copy of the GNU General Public License along with @p0
 import { getHelpMessage } from "../drivers/config";
 import { print1, print2 } from "../drivers/stdio";
 import { checkVersion } from "../middlewares/version";
+import { P0_VERSION_INFO } from "../version";
 import { allowCommand } from "./allow";
 import { awsCommand } from "./aws";
 import { grantCommand } from "./grant";
@@ -42,8 +43,9 @@ const commands = [
   kubeconfigCommand,
 ];
 
-const buildArgv = () => {
-  const argv = yargs(hideBin(process.argv));
+const buildArgv = async () => {
+  const { version } = await P0_VERSION_INFO;
+  const argv = yargs(hideBin(process.argv)).version(version);
 
   // Override the default yargs showHelp() function to include a custom help message at the end
   const originalShowHelp = argv.showHelp.bind(argv);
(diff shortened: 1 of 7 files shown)


src/commands/index.ts
```diff
@@ -42,8 +43,9 @@ const commands = [
   kubeconfigCommand,
 ];
 
-const buildArgv = () => {
-  const argv = yargs(hideBin(process.argv));
+const buildArgv = async () => {
+  const { version } = await P0_VERSION_INFO;
+  const argv = yargs(hideBin(process.argv)).version(version);
```
dhing1024, line 48: This manually tells yargs what the version number is. This was broken in the standalone CLI, because yargs reads this info from the package.json file, which doesn't exist in a standalone installation.

src/middlewares/version.ts
```diff
@@ -43,11 +45,7 @@ export const checkVersion = async (_yargs: yargs.ArgumentsCamelCase) => {
     // Write the version-check file first to avoid retrying errors
     await fs.writeFile(latestFile, "");
 
-    // Note that package.json is installed one level above "dist"
-    // We can't require package.json as it is outside the TypeScript root
-    const { name, version } = JSON.parse(
-      (await fs.readFile(`${__dirname}/../../package.json`)).toString("utf-8")
```
dhing1024, line 49: I think this may have been broken since 0.17.1. Users on 0.17.1 to the current version may not ever see the upgrade message, since this is reading the file at the wrong location.

They may be "stuck" where they are since they have no way to know that the version check failed (the catch block below doesn't do anything).

src/commands/index.ts
```diff
@@ -78,18 +80,22 @@ function conditionalCheckVersion(argv: yargs.ArgumentsCamelCase) {
...
+        print2(error);
+      } else {
+        print2(yargs.help());
+        print2(`\n${message}`);
+        print2(`\n${getHelpMessage()}`);
+      }
+      sys.exit(1);
+    });
+
+  return cli;
```
fabgo, line 100: Nit: can inline this probably.

src/middlewares/version.ts
```diff
@@ -56,15 +54,28 @@ export const checkVersion = async (_yargs: yargs.ArgumentsCamelCase) => {
...
-      print2(
-        `╔══════════════════════════════════════╗
+      if (isSea()) {
+        print2(
+          `╔══════════════════════════════════════╗
+║ A new version of P0 CLI is available ║
+║                                      ║
+║ To install, please view our GitHub   ║
+║ repository README for instructions:  ║
+║ https://github.com/p0-security/p0cli ║
```
fabgo, line 64: Can we just point people to the release page here?

src/drivers/api.ts
```diff
@@ -73,14 +73,15 @@ export const baseFetch = async <T>(
   body: string
 ) => {
   const token = await authn.userCredential.user.getIdToken();
+  const { version } = await P0_VERSION_INFO;
```
fabgo, line 76: A little strange here to await what looks like a constant value.

dhing1024: I renamed this to `p0VersionInfo`

PR #224, opened 2025-07-28
ENG-5460 remove firestore dependency for cli

varunalla:
**Problem**

Current implementation of CLI depends on firestore access for
1. reading configuration
2. listen to request status changes.

This creates firestore stickiness and limits the deployment to gcloud. We might need to strip away this dependency to move the datastore and cloud provider in the backend.

**Change**
1. removes firestore connection from the CLI.
2. uses /orgs endpoint for fetching org data to understand login.
3. uses integration config endpoint to fetch configuration for each of the resource integrations
4. uses request streaming option for requests that need to wait for the request state to transition to resolved status (DONE, DENIED OR ERRORED statuses)

**Validation**
1. updated all unit tests and snapshots
2. manual testing for SSH, request and ls.

**Tracking**
ENG-5460

**Attachments**


src/drivers/org.ts
```diff
@@ -8,7 +8,17 @@ This file is part of @p0security/cli
...
+import { fetchOrgData } from "./api";
+import { print2 } from "./stdio";
 
-export const mockGetDoc = (data: any) =>
-  (getDoc as jest.Mock).mockResolvedValue({ data: () => data });
+export const getOrgData = async (orgId: string) => {
+  try {
+    return await fetchOrgData<RawOrgData>(orgId);
+  } catch (e) {
+    print2(
```
fabgo, line 19: Do we need to disambiguate here? Depending on the status code different things could be going on, e.g. if a 500 error is returned.

src/commands/request.ts
```diff
@@ -17,5 +16,7 @@ export const requestCommand = (yargs: yargs.Argv) =>
     "request [arguments..]",
     "Manually request permissions on a resource",
     requestArgs,
-    fsShutdownGuard(request("request"))
+    async (args) => {
+      await request("request")(args);
```
fabgo, line 20: Consider renaming the "request" function to something a little more meaningful.

src/commands/__tests__/ssh.test.ts
```diff
@@ -61,94 +67,93 @@ const MOCK_REQUEST = {
...
-        status: "APPROVED",
-      });
       const wait = sleep(100);
       await Promise.race([wait, promise]);
       await expect(wait).resolves.toBeUndefined();
+      await expect(promise).resolves.toBeDefined();
     });
-
     it("should call sshOrScp with non-interactive command", async () => {
+      mockStreaming(isPersistent);
```
fabgo, line 144: Could use some code comments to explain this test.

src/commands/shared/request.ts
```diff
@@ -122,41 +100,98 @@ export const request =
...
-      const { id } = data;
-      if (args.wait && id && userCredential.user.tenantId) {
-        const code = await waitForRequest(
-          userCredential.user.tenantId,
-          id,
-          logMessage
-        );
-        if (code) {
-          sys.exit(code);
+    const requestForCommand = async (
```
fabgo, line 103: Why is this named requestForCommand? Consider renaming.

src/commands/shared/request.ts
```diff
@@ -122,41 +100,98 @@ export const request =
...
+      return await logMessageAndProcessData(
+        await requestForCommand(fetchValue()),
+        async (data, logMessage) => {
+          if (logMessage) {
+            print2(data.message);
+            print2("Will wait up to 5 minutes for this request to complete...");
+          }
+
+          for await (const chunkData of fetchStreamingCommandGenerator) {
+            if (!chunkData)
```
fabgo, line 163: Missing { }

src/commands/shared/request.ts
```diff
@@ -122,41 +100,98 @@ export const request =
...
+      return logMessageAndProcessData(
+        await requestForCommand(fetchCommandPromise),
+        async (data, logMessage) => {
+          if (logMessage) print2(data.message);
+          return data;
+        }
+      );
+    };
+    const streamRequest = async () => {
+      const fetchStreamingCommandGenerator = fetchStreamingCommand<
```
fabgo, line 144: Consider renaming this.

src/drivers/api.ts
```diff
@@ -25,50 +29,158 @@ const commandUrl = (tenant: string) => `${tenantUrl(tenant)}/command/`;
 const adminLsCommandUrl = (tenant: string) => `${tenantUrl(tenant)}/command/ls`;
 export const tracesUrl = (tenant: string) => `${tenantUrl(tenant)}/traces`;
 
+export const fetchOrgData = async <T>(orgId: string) =>
+  apiFetch<T>(tenantOrgUrl(orgId), "GET");
+
+export const fetchAccountInformation = async <T>(authn: Authn) =>
```
fabgo, line 35: Nit: Could shorten to fetchAccountInfo

src/drivers/api.ts
```diff
@@ -25,50 +29,158 @@ const commandUrl = (tenant: string) => `${tenantUrl(tenant)}/command/`;
...
+  baseFetch<T>(authn, {
+    url: `${tenantUrl(authn.identity.org.slug)}/account`,
+    method: "GET",
+  });
+
+export const fetchPermissionRequest = async <T>(
```
fabgo, line 41: What does this return? Maybe name it fetchRequestDetails, assuming that is what is being returned?

src/drivers/api.ts
```diff
@@ -25,50 +29,158 @@ const commandUrl = (tenant: string) => `${tenantUrl(tenant)}/command/`;
...
+    method: "POST",
+    body: JSON.stringify({
       requestId: args.requestId,
       publicKey: args.publicKey,
-    })
-  );
+    }),
+  });
+
+export const streamingApiFetch = async function* <T>(
```
fabgo, line 118: Do we need "api" in the name? Rename this to fetchWithStreaming maybe?

src/drivers/api.ts
```diff
@@ -109,21 +220,64 @@ export const auditSshSessionActivity = async (args: {
...
+  } catch (error) {
+    if (error instanceof TypeError && error.message === "fetch failed") {
+      throw `Network error: Unable to reach the server at ${url}.`;
+    } else {
+      throw error;
+    }
+  }
+};
 
+export const apiFetch = async <T>(
```
fabgo, line 265: What's the difference between baseFetch, fetch and apiFetch?

fabgo (commented):
It's a little challenging to understand what is going on. Consider renaming some of the function names to help make the code self-documentating.

src/commands/shared/request.ts
```diff
@@ -65,39 +60,23 @@ export const requestArgs = <T>(yargs: yargs.Argv<T>) =>
...
+  const { status } = permission;
+  if (isCompletedStatus(status)) {
+    const { message, code } = COMPLETED_REQUEST_STATUSES[status];
+    const errorMessage = permission.error
+      ? `${message}: ${permission.error.message}`
+      : message;
+    if (code !== 0 || logMessage) print2(errorMessage);
+    return code;
+  } else {
+    print2("Your request did not complete within 5 minutes.");
```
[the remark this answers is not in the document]
varunalla: yes. at this point the permission should have already transitioned to a resolved state, i.e, DONE,DENIED,ERRORED states

gergas3 (commented):
I'm not sure the logic around `fetchWithStreaming` needs to have that many indirections and complexity. It would good to take a look if it can be simplified.

src/commands/shared/request.ts
```diff
@@ -123,37 +100,98 @@ export const request =
...
+          return data;
+        }
+      );
+    };
+    const executeStreamingRequest = async () => {
+      const fetchStreamingCommandGenerator = fetchStreamingCommand<
+        RequestResponse<T>
+      >(resolvedAuthn, args, [command, ...args.arguments]);
+      const getNextPermissionRequestChunk = async () => {
+        const generatedValue = await fetchStreamingCommandGenerator.next();
```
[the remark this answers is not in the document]
varunalla: yes, the streaming command generator provides the initial request state of NEW, which we use to print the request url to the user. then we wait for changes and let it transition to "DONE".

src/commands/shared/request.ts
```diff
@@ -123,37 +100,98 @@ export const request =
...
+      return await (!args.wait ? invokeRequest() : executeStreamingRequest());
+    } catch (error: any) {
+      if (error instanceof Error && error.name === "TimeoutError") {
+        print2("Your request did not complete within 5 minutes.");
+      }
+      if (
+        error instanceof Error &&
+        (error.name === "disconnect" || error.name === "terminated")
+      ) {
+        print2("Disconnected from server. Retry after 5 minutes.");
```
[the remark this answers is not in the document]
varunalla: if there was connection disruption and the request was already created, we might want the user to retry after sometime and get access, ex: ssh request errored when granting. instead of bombarding the system, they could just wait for a minute and retry it. we can reduce the time though.

src/commands/shared/request.ts
```diff
@@ -123,37 +100,98 @@ export const request =
...
+            }
+            const code = resolveCode(
+              chunkData.request as PermissionRequest<PluginRequest>,
+              logMessage
+            );
+            if (code) {
+              sys.exit(code);
+              return undefined;
+            }
+            return chunkData;
```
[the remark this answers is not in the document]
varunalla: the idea is to surface the first request created url to user and perform async iteration until we reach the generator responds with a valid status.
steps:
1. first response contains the created requestID -> URL presented to user.
2. heartbeats sent from the backend -> skip
3. process response request object sent from backend -> process the code (DENIED,DONE,ERRORED)
4. return undefined or object.

gergas3 (approved):

PR #225, opened 2025-07-30
ENG 5461: Add auth passthrough mechanism

sweeneytr:
Adds a method for passing through the provider access_token to the API calls made, skipping firebase auth.

**Note:** firebase is still initialized + used for org discovery.

### DIFF
--- a/src/drivers/api.ts
+++ b/src/drivers/api.ts
@@ -113,14 +113,13 @@ export const baseFetch = async <T>(
   method: string,
   body: string
 ) => {
-  const token = await authn.userCredential.user.getIdToken();
   const { version } = p0VersionInfo;
 
   try {
     const response = await fetch(url, {
       method,
       headers: {
-        authorization: `Bearer ${token}`,
+        authorization: `Bearer ${await authn.getToken()}`,
         "Content-Type": "application/json",
         "User-Agent": `P0 CLI/${version}`,
       },
--- a/src/drivers/api.ts
+++ b/src/drivers/api.ts
@@ -114,12 +114,13 @@ export const baseFetch = async <T>(
   body: string
 ) => {
   const { version } = p0VersionInfo;
+  const token = await authn.getToken();
 
   try {
     const response = await fetch(url, {
       method,
       headers: {
-        authorization: `Bearer ${await authn.getToken()}`,
+        authorization: `Bearer ${token}`,
         "Content-Type": "application/json",
         "User-Agent": `P0 CLI/${version}`,
       },
--- a/src/drivers/auth/index.ts
+++ b/src/drivers/auth/index.ts
@@ -151,35 +151,35 @@ export const deleteIdentity = async () => {
 };
 
 /** Set up trace exporter to authenticated collector endpoint */
-const setOpentelemetryExporter = async (authn: Authn): Promise<Authn> => {
+const setOpentelemetryExporter = async (authn: Authn): Promise<void> => {
   const url = tracesUrl(authn.identity.org.slug);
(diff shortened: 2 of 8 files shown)


src/drivers/api.ts
```diff
@@ -113,14 +113,13 @@ export const baseFetch = async <T>(
...
 ) => {
-  const token = await authn.userCredential.user.getIdToken();
   const { version } = p0VersionInfo;
 
   try {
     const response = await fetch(url, {
       method,
       headers: {
-        authorization: `Bearer ${token}`,
+        authorization: `Bearer ${await authn.getToken()}`,
```
fabgo, line 122: Nit: keep the token variable.

src/drivers/auth/index.ts
```diff
@@ -152,26 +151,35 @@ export const deleteIdentity = async () => {
...
 
 /** Set up trace exporter to authenticated collector endpoint */
-const setOpentelemetryExporter = async (
-  identity: Identity,
-  userCredential: UserCredential
-) => {
-  const idToken = await userCredential.user.getIdToken();
-  const url = tracesUrl(identity.org.slug);
-  await setExporterAfterLogin(url, idToken);
+const setOpentelemetryExporter = async (authn: Authn): Promise<Authn> => {
```
fabgo, line 154: I think it would be useful to have a more specific name than authn/Authn. Consider renaming.

src/drivers/auth/index.ts
```diff
@@ -152,26 +151,35 @@ export const deleteIdentity = async () => {
...
+        getToken: () => Promise.resolve(identity.credential.access_token),
+      };
+    }
 
-  await setOpentelemetryExporter(identity, userCredential);
+    // Note: if the `providerId` is "password", we already actually already
+    // retrieved the UserCredential object in `loadCredentialsWithAutoLogin`.
+    // This following call to `authenticateToFirebase` could be omitted.
+    const userCredential = await authenticateToFirebase(identity, options);
+    return {
```
fabgo, line 177: Define a type for this?

src/drivers/auth/index.ts
```diff
@@ -152,26 +151,35 @@ export const deleteIdentity = async () => {
...
+      identity,
+      userCredential,
+      getToken: () => userCredential.user.getIdToken(),
+    };
+  };
 
-  return { userCredential, identity };
+  return await getAuthn().then(setOpentelemetryExporter);
```
fabgo, line 184: I am curious about this pattern. Why not just invoke it after authenticateToFirebase and avoid the inner function?

src/drivers/auth/index.ts
```diff
@@ -152,26 +151,35 @@ export const deleteIdentity = async () => {
...
   debug?: boolean;
 }): Promise<Authn> => {
-  const identity = await loadCredentialsWithAutoLogin(options);
-  // Note: if the `providerId` is "password", we already actually already
-  // retrieved the UserCredential object in `loadCredentialsWithAutoLogin`.
-  // This following call to `authenticateToFirebase` could be omitted.
-  const userCredential = await authenticateToFirebase(identity, options);
+  const getAuthn = async () => {
+    const identity = await loadCredentialsWithAutoLogin(options);
+    if (identity.org.authPassthrough) {
```
fabgo, line 166: What is this authPassthrough? This deserves a clarifying comment I think.

sweeneytr: Added a JSDoc comment to the field, and renamed it the clearer `useProviderToken`

src/drivers/auth/index.ts
fabgo: OK

PR #229, opened 2025-08-06
Fix SSH on Windows

dhing1024:


dhing1024:
Closing this because this solution seems to not work very reliably (potentially different failure modes on different Windows OSes and terminals)

PR #230, opened 2025-08-11
ENG-5443 Add bootstrap org config

fabgo:
Adds the ability to configure the P0 CLI using an optional `bootstrap-<org id>.json` file that gets placed in the `~\.p0` directory. This alleviates the need to connect to P0 servers to load the org config.


bgberr (approved):
One comment. Otherwise looks good

PR #232, opened 2025-08-12
Fix intermittent failures using p0 ssh

varunalla:
**Problem**

P0 cli faces intermittent issues because of streaming response having jsonl format. The client tries to parse jsonl and fails. i.e, if the server responds with more than one json separated by new lines in a single chunk.

**Change**

Update client streaming method to parse response as jsonl.

**Implementation**
1. instead of processing response as a single object, break on new line and parse them as individual strings.
2. Reuse jsonlToArray implementation.

### DIFF
--- a/src/drivers/__tests__/api.test.ts
+++ b/src/drivers/__tests__/api.test.ts
@@ -317,7 +317,7 @@ describe("fetchWithStreaming", () => {
       results.push(chunk);
     }
 
-    expect(results.length).toBeGreaterThanOrEqual(0);
+    expect(results.length).toEqual(0);
   });
 
   it("should handle empty chunks", async () => {
--- a/src/drivers/api.ts
+++ b/src/drivers/api.ts
@@ -147,7 +147,7 @@ export const fetchWithStreaming = async function* <T>(
     );
 
     if (!response.body) throw "No reader available";
-    const onLine = async function* (line: string) {
+    const onLine = (line: string) => {
       const segment = JSON.parse(line);
       if (segment.type === "error") {
         throw segment.error;
@@ -160,8 +160,9 @@ export const fetchWithStreaming = async function* <T>(
         if ("error" in data) {
           throw data.error;
         }
-        yield data as T;
+        return data as T;
       }
+      return undefined; // Ignore heartbeat messages
     };
     // we need get the reader from the body as the backend will be streaming chunks of stringified json
     // response data delimited using new lines.
@@ -187,7 +188,10 @@ export const fetchWithStreaming = async function* <T>(
       buffer = parts.pop() ?? "";
 
       for (const line of parts) {
-        yield* onLine(line);
+        const response = onLine(line);
+        if (response) {
+          yield response;
+        }
       }
     }
     // do not handle the left over buffer as it may contain partial json and the backend is always expected to send complete json objects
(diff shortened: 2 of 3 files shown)


src/drivers/api.ts
```diff
@@ -154,22 +154,27 @@ export const fetchWithStreaming = async function* <T>(
         break;
       }
       const value = read.value;
-      const text = textDecoder.decode(value);
-      const parsedResponse = JSON.parse(text);
-      if (parsedResponse.type === "error") {
-        throw new Error(parsedResponse.error);
+      const jsonlSegments = convertJsonlToArray<{
```
dhing1024, line 157: I'm not sure this fully solves the problem. I think it relies on the assumption that when reading from the reader(), the individual chunks that the reader gives you are going to be auto-chunked into jsonl segments, but I think the chunking can be arbitrary.

I was running into this chunking issue when working on other CLI stuff (I was on a Windows VM, but that may not be relevant). I can try to see if I can repro that.

src/drivers/__tests__/api.test.ts
```diff
@@ -0,0 +1,443 @@
...
+
+    const results = [];
+    for await (const chunk of generator) {
+      results.push(chunk);
+    }
+
+    expect(results).toEqual([{ id: "1\ntest" }, { id: "2" }]);
+  });
+
+  it("should handle chunks with partial newlines", async () => {
```
fabgo, line 227: Partial newlines? Do you mean paertial data?

src/drivers/__tests__/api.test.ts
```diff
@@ -0,0 +1,443 @@
...
+    });
+
+    await expect(async () => {
+      for await (const _chunk of generator) {
+        // Should throw before yielding
+      }
+    }).rejects.toBe("Invalid response from the server");
+  });
+
+  it("should handle multiple JSON objects in single chunk", async () => {
```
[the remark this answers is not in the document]
varunalla: these are different scenarios. line 62 assumes that the backend sends two different chunks separated in time. but line 203 assumes that backend sends two jsons in a single chunk separated by new lines.

src/drivers/__tests__/api.test.ts
```diff
@@ -223,21 +224,125 @@ describe("fetchWithStreaming", () => {
...
+      url: "/stream",
+      method: "GET",
+    });
+
+    const results = [];
+    for await (const chunk of generator) {
+      results.push(chunk);
+    }
+
+    expect(results.length).toBeGreaterThanOrEqual(0);
```
dhing1024, line 320: Did you mean for this to be `expect(results).toEqual(...)`?

src/drivers/api.ts
```diff
@@ -146,26 +145,14 @@ export const fetchWithStreaming = async function* <T>(
...
     );
-    const reader = response.body?.getReader();
-    if (!reader) throw `No reader available`;
-    while (true) {
-      const read = await reader.read();
-      if (read.done) {
-        break;
+
+    if (!response.body) throw "No reader available";
+    const onLine = async function* (line: string) {
```
dhing1024, line 150: Clarification: does `onLine` here need to be a generator function?

dhing1024 (approved):

GGonryun:
How does this handle errors? I had to write something similar to get my CLI to work with on-prem ssh and realized that if we timeout or if we return an error from the backend it gets rendered funny. 🤔

fabgo (approved):

PR #233, opened 2025-08-13
Fix gcloud resolution on Windows

dhing1024:
Fixes gcloud SSH on Windows machines. The problem is that on Windows, when installing the Google Cloud tools, the main `gcloud` file is a `.cmd` (shell script) file rather than a `.exe` (binary executable) file, so when calling `spawn`, it cannot be located.

That is because the underlying Windows OS API that `spawn` uses doesn't resolve `.CMD` files by default (MacOS does not have this problem because MacOS will automatically resolve "gcloud" to the correct executable).

Fix is when on a Windows machine to instead call `cmd.exe` with `gcloud.cmd` and the remaining parameters as args, which will resolve the correct file.

Validated that `gcloud` could be correctly resolved on:
- [x] Parallels desktop pro for MacOS
- [x] Azure VM (Windows 11 23h2 client)
- [x] Azure VM (Windows 11 24h2 client)

### DIFF
--- a/src/plugins/google/ssh-key.ts
+++ b/src/plugins/google/ssh-key.ts
@@ -10,6 +10,7 @@ You should have received a copy of the GNU General Public License along with @p0
 **/
 import { asyncSpawn } from "../../common/subprocess";
 import { print2 } from "../../drivers/stdio";
+import { getOperatingSystem } from "../../util";
 import { ImportSshPublicKeyResponse } from "./types";
 
 /**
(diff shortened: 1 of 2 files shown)


src/plugins/google/ssh-key.ts
```diff
@@ -28,17 +29,25 @@ export const importSshKey = async (
   options?: { debug?: boolean }
 ) => {
   const debug = options?.debug ?? false;
+  const isWindows = getOperatingSystem() === "win";
```
fabgo, line 32: Maybe this should be built into asyncSpan instead?

dhing1024: I think at least for now it makes sense to keep `asyncSpawn` at little more generic.

PR #237, opened 2025-08-15
Fix stream api handling of errors without type and new line

varunalla:
Problem:

We do not handle stream api errors which fail during permission request creation. This is apparent when we try to ssh into a non-existent node.
1. the errors thrown by errorBoundary and unhandled error objects from the backend do not have "type" defined like streaming responses.
2. these errors are not new line delimited.

Change:
1. introduce way to debug these requests.
2. try parsing the leftover buffer and validate compelete-ness.
3. if the json is error, surface the error to the user.

Tracking:
ENG-5801

Commands:
the node should not exist
./p0 ssh dummy-node-123
./p0 ssh dummy-node-123 --debug

Attachements:

Before:

image

After:

Screenshot 2025-08-15 at 11 50 09 AM

Screenshot 2025-08-15 at 11 50 28 AM

### DIFF
--- a/src/drivers/api.ts
+++ b/src/drivers/api.ts
@@ -202,33 +202,39 @@ export const fetchWithStreaming = async function* <T>(
     }
     // do not handle the left over buffer as it may contain partial json and the backend is always expected to send complete json objects
     if (buffer.length > 0) {
-      // log the error for reference
-      // this should not happen in most scenarios
-      if (debug)
+      // this should not happen in most scenarios except errors
+      if (debug) {
         print2(
-          "[API:stream]Incomplete data received from the server: " + buffer
+          "[API:stream]Remaining data received from the server but not processed due to the lack of new-line: " +
+            buffer
         );
-      // try handle non streaming api chunks - error boundary messages
+      }
+      // there is a chance that the server could have errored with a non-streaming response
+      // we hit a errorBoundary in the backend and we received a valid json without a new-line delimiter at the end
       try {
-        if (debug)
+        if (debug) {
           print2(
             "[API:stream]Trying to parse to validate json completeness: " +
               buffer
           );
+        }
         const response = JSON.parse(buffer);
         if ("error" in response) {
           throw response.error;
         }
       } catch (err) {
-        // if this is json parse error consume it
+        // If this is a json parse error then we have received a partial response
+        // we could throw an error saying incomplete response from the server
         // else rethrow the error
         if (err instanceof SyntaxError) {
-          // log the error for reference
-          if (debug)
+          // log the error in debug logs
+          if (debug) {
             print2(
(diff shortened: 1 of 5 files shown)


src/drivers/api.ts
```diff
@@ -198,7 +204,35 @@ export const fetchWithStreaming = async function* <T>(
     if (buffer.length > 0) {
       // log the error for reference
       // this should not happen in most scenarios
-      print2("Incomplete data received from the server" + buffer);
+      if (debug)
```
fabgo, line 207: Use {} with multi-line if statements. This it to prevent bugs like:

if (condition)
   do something;
   do something else;

src/drivers/api.ts
```diff
@@ -198,7 +204,35 @@ export const fetchWithStreaming = async function* <T>(
     if (buffer.length > 0) {
       // log the error for reference
```
fabgo, line 205: Comments may need an update. It's no longer necessarily an error, as I understand it?

src/drivers/api.ts
```diff
@@ -198,7 +204,35 @@ export const fetchWithStreaming = async function* <T>(
...
+        const response = JSON.parse(buffer);
+        if ("error" in response) {
+          throw response.error;
+        }
+      } catch (err) {
+        // if this is json parse error consume it
+        // else rethrow the error
+        if (err instanceof SyntaxError) {
+          // log the error for reference
+          if (debug)
```
fabgo, line 227: Not sure we need to log here if we throw the error.

src/drivers/api.ts
```diff
@@ -198,7 +204,35 @@ export const fetchWithStreaming = async function* <T>(
     if (buffer.length > 0) {
       // log the error for reference
       // this should not happen in most scenarios
-      print2("Incomplete data received from the server" + buffer);
+      if (debug)
+        print2(
+          "[API:stream]Incomplete data received from the server: " + buffer
```
fabgo, line 209: Why is this data incomplete? If it's a valid JSON, isn't it by definition complete?

fabgo (commented):
I have some questions...

src/drivers/api.ts
```diff
@@ -198,7 +204,35 @@ export const fetchWithStreaming = async function* <T>(
...
+        );
+      // try handle non streaming api chunks - error boundary messages
```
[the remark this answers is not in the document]
varunalla: updated the comment to add why we need this logic.

src/drivers/api.ts
varunalla: fixed the messaging

src/drivers/api.ts
varunalla: updated the logic to log and re-throw it.

PR #240, opened 2025-08-25
Switch to certificates for self-hosted SSH access

GGonryun:
**Problem**

Public keys are not as secure as certificates for SSH access.

**Change**

Implements automatic registration and short-lived certificate access.

**Validation**

1. Registration Process & changing a node's name.
image

image

image

2. Making an access request.
image

3. Connecting with tailscale on NixOS
image

**Tracking**

https://linear.app/p0-security/issue/ENG-5826/switch-to-short-lived-certificates-for-access

**Implementation**

1. We switched to automatic registration so that user's don't need to manually configure the node anymore, they just pass in the api key (which will switch to auth token)

2. We're saving the CA public key in the user's authorized_keys file, this makes it so that machines don't implicitly trust any signed certificate but allows us to selectively determine which machines a user has access to without being in-band using the `AuthorizedPrincipalsCommand`. 

3. One other advantage of this approach is that rotating CA's is much easier since with every access request we'll only insert the most up to date certificate when access is provisioned. There is one edge case where this might not be true is if the CA rotates while a user has access, they will need to re-request but this can be solved later. 

4. If you are developing using ngrok you will need to modify your tunnelHost in your config in order to connect to your braekhus proxy.

**Future Work**
https://linear.app/p0-security/issue/ENG-5895/add-support-for-automatic-ca-rotation
https://linear.app/p0-security/issue/ENG-5896/restrict-certificates-to-a-single-target-machine
https://linear.app/p0-security/issue/ENG-5897/add-a-watch-dog-service-to-terminate-potentially-expired-access
https://linear.app/p0-security/issue/ENG-5898/add-a-ephemeral-user-toggle-to-self-hosted-environment-config
https://linear.app/p0-security/issue/ENG-5899/execute-a-force-command-on-ssh-login-that-places-users-in-a-nix-shell

### DIFF
--- a/src/drivers/api.ts
+++ b/src/drivers/api.ts
@@ -13,7 +13,6 @@ import { p0VersionInfo } from "../version";
 import { getTenantConfig } from "./config";
 import { defaultConfig } from "./env";
 import { print2 } from "./stdio";
-import { pick } from "lodash";
 import * as path from "node:path";
 import yargs from "yargs";
 
@@ -124,15 +123,17 @@ export const submitPublicKey = async <T>(
     }),
   });
 
-const CSR_KEYS = ["requestId", "publicKey"] as const;
 export const certificateSigningRequest = async (
   authn: Authn,
-  args: Record<(typeof CSR_KEYS)[number], string>
+  args: { publicKey: string; requestId: string }
 ) =>
   baseFetch<{ signedCertificate: string }>(authn, {
     url: certSignRequestUrl(authn.identity.org.slug),
     method: "POST",
-    body: JSON.stringify(pick(args, CSR_KEYS)),
+    body: JSON.stringify({
+      requestId: args.requestId,
+      publicKey: args.publicKey,
+    }),
   });
 
 export const fetchWithStreaming = async function* <T>(
--- a/src/drivers/api.ts
+++ b/src/drivers/api.ts
@@ -23,6 +23,8 @@ const tenantOrgUrl = (tenant: string) =>
 const tenantUrl = (tenant: string) => `${getTenantConfig().appUrl}/o/${tenant}`;
 const publicKeysUrl = (tenant: string) =>
   `${tenantUrl(tenant)}/integrations/ssh/public-keys`;
+const certSignRequestUrl = (tenant: string) =>
+  `${tenantUrl(tenant)}/integrations/ssh/certificates`;
 const sshAuditUrl = (tenant: string) =>
   `${tenantUrl(tenant)}/integrations/ssh/audit`;
 
@@ -114,6 +116,19 @@ export const submitPublicKey = async <T>(
 ) =>
   authFetch<T>(authn, {
     url: publicKeysUrl(authn.identity.org.slug),
+    method: "POST",
+    body: JSON.stringify({
+      requestId: args.requestId,
+      publicKey: args.publicKey,
+    }),
+  });
+
+export const certificateSigningRequest = async (
+  authn: Authn,
+  args: { publicKey: string; requestId: string }
+) =>
+  baseFetch<{ signedCertificate: string }>(authn, {
+    url: certSignRequestUrl(authn.identity.org.slug),
     method: "POST",
     body: JSON.stringify({
       requestId: args.requestId,
--- a/src/drivers/api.ts
+++ b/src/drivers/api.ts
@@ -127,7 +127,7 @@ export const certificateSigningRequest = async (
   authn: Authn,
   args: { publicKey: string; requestId: string }
 ) =>
-  baseFetch<{ signedCertificate: string }>(authn, {
+  authFetch<{ signedCertificate: string }>(authn, {
     url: certSignRequestUrl(authn.identity.org.slug),
     method: "POST",
     body: JSON.stringify({
--- a/src/drivers/api.ts
+++ b/src/drivers/api.ts
@@ -13,6 +13,7 @@ import { p0VersionInfo } from "../version";
 import { getTenantConfig } from "./config";
 import { defaultConfig } from "./env";
 import { print2 } from "./stdio";
+import { pick } from "lodash";
 import * as path from "node:path";
 import yargs from "yargs";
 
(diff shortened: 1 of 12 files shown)


src/drivers/api.ts
```diff
@@ -121,6 +123,19 @@ export const submitPublicKey = async <T>(
...
+  args: { publicKey: string; requestId: string }
+) =>
+  baseFetch<{ signedCertificate: string }>(authn, {
+    url: certSignRequestUrl(authn.identity.org.slug),
+    method: "POST",
+    body: JSON.stringify({
+      requestId: args.requestId,
+      publicKey: args.publicKey,
+    }),
+  });
```
sweeneytr, line 137: **golf** If this was going to change frequently, or define many keys to forward, the below would be a smooth way to do it.

```suggestion
const CSRKeys = ["requestId", "publicKey"] as const;
export const certificateSigningRequest = async (
  authn: Authn,
  args: Record<CSRKeys[number], string>
) =>
  baseFetch<{ signedCertificate: string }>(authn, {
    url: certSignRequestUrl(authn.identity.org.slug),
    method: "POST",
    body: JSON.stringify(pick(args, CSRKeys)),
  });
```

src/plugins/ssh/shared.ts
```diff
@@ -0,0 +1,26 @@
...
+
+@p0security/cli is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 3 of the License.
+
+@p0security/cli is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
+**/
+import tmp from "tmp-promise";
+
+export const createTempDirectoryForKeys = async (): Promise<{
```
sweeneytr, line 13: **remark** I've seen tooling like this that creates temporary directories break down in more secure environments where there's no writable tmp directory mounted.

I think we should leave it as-is, but **far** in the future we may need to revisit this & make the tmp dir behavior configurable.

src/plugins/self-hosted/keygen.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@p0security/cli is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
+**/
+import { certificateSigningRequest } from "../../drivers/api";
+import { print2 } from "../../drivers/stdio";
+import { Authn } from "../../types/identity";
+
+export const generateSelfHostedCertificate = async (
```
sweeneytr, line 15: **nit** at first (and second) read this sounds like `SelfSigned`. You might want to come up with a more distinct name, I feel like at docs / communication time this could become a problem (maybe `generateAgentCertificate`?).

src/plugins/self-hosted/ssh.ts
```diff
@@ -51,9 +57,43 @@ export const selfHostedSshProvider: SshProvider<
...
+    await fs.writeFile(sshCertificateKeyPath, signedCertificate);
+
+    const teardown = async () => {
+      await sshKeyPathCleanup();
+    };
+
     return {
-      privateKeyPath: PRIVATE_KEY_PATH,
+      sshOptions: [`CertificateFile=${sshCertificateKeyPath}`],
+      teardown,
```
sweeneytr, line 96: **nit** teardown` is a no-op helper, you could delete it above + return the callable directly
```suggestion
      teardown: sshKeyPathCleanup
```

src/plugins/self-hosted/ssh.ts
```diff
@@ -51,9 +57,43 @@ export const selfHostedSshProvider: SshProvider<
...
+  setup: async (authn, _request, options) => {
+    const { path: keyPath, cleanup: sshKeyPathCleanup } =
+      await createTempDirectoryForKeys();
+    const { publicKey } = await createKeyPair();
+
+    const signedCertificate = await generateSelfHostedCertificate(authn, {
+      debug: options.debug,
+      requestId: options.requestId,
+      publicKey,
+    });
```
sweeneytr, line 86: **nit** Here and a few other spots, you could leverage spreads to minimize rebinding of key-vals
```suggestion
    const signedCertificate = await generateSelfHostedCertificate(authn, {
      ...options,
      publicKey,
    });
```

src/plugins/ssh/shared.ts
sweeneytr: Update - given how it's used, you might want to explore writing to memory, since the tmp files are only needed for upload.

src/plugins/ssh/shared.ts
```diff
@@ -0,0 +1,26 @@
...
+  // unsafeCleanup lets us delete the directory even if there are still files in it, which is fine since the
+  // files are no longer needed once we've authenticated to the remote system.
+  const { path, cleanup } = await tmp.dir({
+    mode: 0o700,
+    prefix: "p0cli-",
+    unsafeCleanup: true,
+  });
+
+  return { path, cleanup };
+};
```
sweeneytr, line 26: **nit** if cleanup was critical (which it likely isn't), you might want to design this as a wrapper like so:

```suggestion
export const withTempDir = async <T>(f: (path: string) => T): Promise<T> => {
  // unsafeCleanup lets us delete the directory even if there are still files in it, which is fine since the
  // files are no longer needed once we've authenticated to the remote system.
  const { path, cleanup } = await tmp.dir({
    mode: 0o700,
    prefix: "p0cli-",
    unsafeCleanup: true,
  });

 try {
   return f(path);
 } finally {
   await cleanup();
 }
};
```

Then, you could use it like:
```ts
await withTempDir((keyDir) => {
  const myKey = path.join(keyDir, MY_KEY_FILENAME);
  ...
})
```

sweeneytr (approved):
All **nits**, at various degrees of actionability / necessity. LGTM, fix up the ones you agree w/ and go ahead + merge.

PR #242, opened 2025-08-26
handle both '--size n' and '--size=n' args

GGonryun:
This PR fixes a bug which prevents the `--size=n` argument from working correctly. Right now this command makes LS fail:
image

After this update both `--size n` and `--size=n` works.
image
image
image

### DIFF
--- a/src/commands/ls.ts
+++ b/src/commands/ls.ts
@@ -63,13 +63,29 @@ export const lsCommand = (yargs: yargs.Argv) =>
  */
 const convertLsSizeArg = (args: string[]) => {
   const convertedArgs = [...args];
+
   const sizeIndex = convertedArgs.findIndex((a) => a === "--size");
-  const requestedSize = +(
-    (sizeIndex >= 0
-      ? pullAt(convertedArgs, sizeIndex, sizeIndex + 1)[1]
-      : undefined) ?? DEFAULT_RESPONSE_SIZE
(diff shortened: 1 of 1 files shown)


src/commands/ls.ts
```diff
@@ -63,13 +63,29 @@ export const lsCommand = (yargs: yargs.Argv) =>
...
   );
+
+  let requestedSize: number;
+
+  if (sizeIndex >= 0) {
+    // Handle --size n format
+    const sizeValue = pullAt(convertedArgs, sizeIndex, sizeIndex + 1)[1];
+    requestedSize = +(sizeValue ?? DEFAULT_RESPONSE_SIZE);
+  } else if (sizeEqualIndex >= 0) {
+    // Handle --size=n format
```
GGonryun, line 79: breaking this up out of a ternary makes it easier to read.

src/commands/ls.ts
```diff
@@ -40,57 +40,56 @@ const lsArgs = <T>(yargs: yargs.Argv<T>) =>
       string: true,
       default: [] as string[],
     })
+    .option("size", {
```
GGonryun, line 43: this lets us intercept the `--size` flag, and re-add it into `arguments` before sending it to the api backend without having to do any special text parsing.

fabgo (approved):

PR #243, opened 2025-08-27
Friendly error message for errors from web application firewall

gergas3:
The WAF of the P0 backend returns error messages as an HTML page.
These error messages were not surfaced to the user, instead the CLI failed parsing them as JSON and surfaced that error.

Changes:
- First check if the response is in the format that the WAF returns: contains a <title> tag with the status code in it. If yes, return that as an error message. If no, continue with parsing as JSON.
- Opportunistic change: refactor `baseFetch` and `unauthenticatedFetch` to use a common base function - which is called `baseFetch`. The net result is that `unauthenticatedFetch` becomes `baseFetch` because authentication is an additional header on top of the base request format. So the old `baseFetch` the becomes `authFetch`.

### DIFF
--- a/src/drivers/api.ts
+++ b/src/drivers/api.ts
@@ -31,10 +31,10 @@ const adminLsCommandUrl = (tenant: string) => `${tenantUrl(tenant)}/command/ls`;
 export const tracesUrl = (tenant: string) => `${tenantUrl(tenant)}/traces`;
 
 export const fetchOrgData = async <T>(orgId: string) =>
-  unauthenticatedApiFetch<T>(tenantOrgUrl(orgId), "GET");
+  baseFetch<T>({ url: tenantOrgUrl(orgId), method: "GET" });
 
 export const fetchAccountInfo = async <T>(authn: Authn) =>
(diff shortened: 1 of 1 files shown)


src/drivers/api.ts
```diff
@@ -333,35 +330,40 @@ export const baseFetch = async <T>(
...
+  if (!match) {
+    return undefined;
   }
+  const statusCode = match[1];
+  const statusText = text
+    // Remove the title tag
+    .replace(/<title>(\d{3})<\/title>/g, "")
+    // Remove meta HTML tags
+    .replace(/<[^>]+>/g, "");
+  return `${statusText}: (HTTP status ${statusCode})`;
```
gergas3, line 368: This function and its use in `baseFetch` is the new addition.

GGonryun (approved):
I feel like our API shouldn't even be returning these errors as html 💀

gergas3:
> I feel like our API shouldn't even be returning these errors as html 💀

It's not our API but the web application firewall - and it can't be configured to return json

PR #248, opened 2025-09-05
AWS GovCloud support for ssh with Identity Center login

gergas3:
This PR adds support for ssh into EC2 instances in an AWS GovCloud account that uses Identity Center.
Does _not_ add GovCloud support for GovCloud accounts that use Federated login.

- The IDC start URL is different for GovCloud
- The partition in the ARN is also different, wildcarded that in the error message we check for when waiting for access propagation

### DIFF
--- a/src/plugins/aws/idc/index.ts
+++ b/src/plugins/aws/idc/index.ts
@@ -63,12 +63,10 @@ const awsIdcHelpers = (
   const { id, region } = idc;
 
   // The start url can be customized with a subdomain. Here only the default is supported.
-  const buildStartUrl = () => {
-    if (region.includes("us-gov")) {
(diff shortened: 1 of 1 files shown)


src/plugins/aws/idc/index.ts
```diff
@@ -62,13 +62,21 @@ const awsIdcHelpers = (
   const { clientId, clientSecret } = clientCredentials;
   const { id, region } = idc;
 
+  // The start url can be customized with a subdomain. Here only the default is supported.
+  const buildStartUrl = () => {
+    if (region.includes("us-gov")) {
```
fabgo, line 67: Nit: consider using ternary operator.

PR #262, opened 2025-09-30
ENG-6129 Correctly handle 400 response

fabgo:
Server returns a 400 response if CLI command is incomplete, 
which is not handled correctly after we added parsing of 429 errors.

* Always try to parse response as JSON first
* If not JSON, fall back to parse HTTP status codes

### DIFF
--- a/src/drivers/api.ts
+++ b/src/drivers/api.ts
@@ -242,7 +242,7 @@ export const fetchWithStreaming = async function* <T>(
               buffer
           );
         }
-        handleResponse(response, buffer);
+        handleResponse(response, buffer, debug);
       } catch (err) {
         // If this is a json parse error then we have received a partial response
         // we could throw an error saying incomplete response from the server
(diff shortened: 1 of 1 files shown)


src/drivers/api.ts
```diff
@@ -382,10 +382,17 @@ const authFetch = async <T>(
...
-    throw `HTTP Error: ${response.status} ${response.statusText}`;
+  let data;
+  try {
+    data = JSON.parse(responseText);
+  } catch (err) {
+    if ("ok" in response && !response.ok) {
+      throw `HTTP Error: ${response.status} ${response.statusText}`;
+    } else {
+      throw "Invalid response from the server";
+    }
```
gergas3, line 393: nit suggestion: when we hit the else branch it will be hard to know what went wrong. Can we log the error and responseText in debug mode? (`debug` is already available in the callers of this function.)
```
const handleResponse = (response: Response, responseText: string, debug?: boolean) => {
  let data;
  try {
    data = JSON.parse(responseText);
  } catch (err) {
    if ("ok" in response && !response.ok) {
      throw `HTTP Error: ${response.status} ${response.statusText}`;
    } else {
      if (debug) {
        print2(`Parse error: ${String(err)}\nin response: ${responseText}`);
      }
      throw "Invalid response from the server";
    }
```

gergas3 (approved):

src/drivers/api.ts
fabgo: Done

PR #263, opened 2025-09-30
Fix CLI double login bug, improve debug logging

dhing1024:
On Windows, the NPM version check was broken, which led to the end user sometimes needing to run the login command twice. This fixes the core issues at play, but also improves debug logging for easier traceability.

- Adjusts the spawn of the NPM version check to account for the fact that NPM isn't a .exe file.
- Adds a .on("error") check to the `exec` implementation to explicitly handle errors (required to prevent uncaught exceptions)
- Adds global process error handlers to improve logging when uncaught errors or unhandled rejections occur (without this change, nothing printed to the terminal; with this change, it would have made it much easier to debug this particular issue)
- Improves overall debug logging of the NPM version checking code paths
- Adjusts `.on("exit")` to `.on("close")` to allow stdio/stderr to cleanly finish
- Bump NPM version check timeout from 1s to 2s to improve reliability (it took generally longer on my local Virtualized Windows hardware - could be either due to the Virtualization or Windows itself being slower)

I tested this on both Windows and MacOS to test the login flows, debug logging, and that SSH functionality still works

### DIFF
--- a/src/drivers/auth/index.ts
+++ b/src/drivers/auth/index.ts
@@ -163,7 +163,9 @@ export const authenticate = async (options?: {
   const identity = await loadCredentialsWithAutoLogin(options);
   if (options?.debug) {
     print2(`Loaded identity for user for org ${identity.org.slug}`);
-    print2(`Token expires in ${remainingTokenTime(identity)} seconds`);
+    print2(
+      `Token expires in ${Math.floor(remainingTokenTime(identity))} seconds`
+    );
   }
   let authn: Authn;
 
--- a/src/drivers/auth/index.ts
+++ b/src/drivers/auth/index.ts
@@ -104,7 +104,7 @@ export const loadCredentials = async (): Promise<Identity> => {
 };
 
 export const remainingTokenTime = (identity: Identity) =>
-  identity.credential.expires_at - Date.now() * 1e-3;
+  Math.floor(identity.credential.expires_at - Date.now() * 1e-3);
 
 const loadCredentialsWithAutoLogin = async (options?: {
   noRefresh?: boolean;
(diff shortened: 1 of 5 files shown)


src/drivers/auth/index.ts
```diff
@@ -163,7 +163,9 @@ export const authenticate = async (options?: {
   const identity = await loadCredentialsWithAutoLogin(options);
   if (options?.debug) {
     print2(`Loaded identity for user for org ${identity.org.slug}`);
-    print2(`Token expires in ${remainingTokenTime(identity)} seconds`);
+    print2(
+      `Token expires in ${Math.floor(remainingTokenTime(identity))} seconds`
```
fabgo, line 167: Nit: Math.floor can go inside remainingTokenTime

src/middlewares/version.ts
```diff
@@ -21,39 +21,79 @@ const LATEST_VERSION_FILE = "last-version-check";
...
+    }
 
-    const npmResult = exec("npm", ["view", name, "--json"], { check: true });
-    const npmPackage = await timeout(npmResult, VERSION_CHECK_TIMEOUT_MILLIS);
-    const {
-      "dist-tags": { latest },
-    } = JSON.parse(npmPackage.stdout);
+    // On Windows, the main npm file is not an .exe (binary executable) file,
+    // so when calling spawn, it cannot be located except via cmd.exe
+    const isWindows = getOperatingSystem() === "win";
```
fabgo, line 76: Isn't this duplicate code? Can this be moved to a wrapper function around exec/spawn to handle this in all cases?

fabgo (approved):

src/middlewares/version.ts
```diff
@@ -21,39 +21,79 @@ const LATEST_VERSION_FILE = "last-version-check";
...
+          );
+        }
         return;
+      }
     } catch (error: any) {
       if (error.code !== "ENOENT") throw error;
     }
 
     // Write the version-check file first to avoid retrying errors
+    // Ensure that the directory exists beforehand
```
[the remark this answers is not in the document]
dhing1024: Root cause was that NPM process to do the version check wouldn't spawn correctly

PR #264, opened 2025-10-01
ENG-6034 Use $0 to identify the name of the executable

fabgo:
- Clean up messaging, using a dynamic name for the executable, instead of a hard-coded `p0`.`


gergas3 (approved):
I searched for any remaining user-visible instances of `"p0"`.
One remaining instance is in `redirect-landing.html`, however, that page only gets rendered with Google and Azure login, and not with Okta login.

PR #265, opened 2025-10-01
Update the P0 CLI version to 0.19.4

fabgo:
Update the P0 CLI version to 0.19.4.


dhing1024:
I saw that you update this line of code:
```
export const getAppPath = () =>
  process.env.P0_APP_PATH ?? process.argv[1] ?? "p0";
```

Does this work if the CLI is the SEA version?

PR #268, opened 2025-10-03
Gcloud SSH: Fallback to local SSH public key for OS Login Import

dhing1024:
This change fixes a bug where in some circumstances, the request object won't contain a public key due to the manner in which it was originally requested. In those situations, the public key won't get imported into the user's Google account, preventing them from SSH-ing into gcloud VMs.

This change resolves this issue by instead using whichever SSH public key is currently on the user's machine (in the `~/.p0/ssh` folder).

Additionally, to improve error handling by failing more quickly, `importSshKey` throws if the public key is missing.

### DIFF
--- a/src/plugins/google/ssh.ts
+++ b/src/plugins/google/ssh.ts
@@ -129,12 +129,7 @@ export const gcpSshProvider: SshProvider<
   toCliRequest: async (request, options) => ({
     ...request,
     cliLocalData: {
-      linuxUserName: await importSshKey(
-        // fallback to the local public key if the one in the request object is missing, which
-        // can happen if, for example, the request was created using the Slack modal or web app
-        request.permission.publicKey ?? options.publicKey,
-        options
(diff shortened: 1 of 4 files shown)


GGonryun (approved):

src/plugins/google/ssh.ts
```diff
@@ -129,7 +129,12 @@ export const gcpSshProvider: SshProvider<
   toCliRequest: async (request, options) => ({
     ...request,
     cliLocalData: {
-      linuxUserName: await importSshKey(request.permission.publicKey, options),
+      linuxUserName: await importSshKey(
+        // fallback to the local public key if the one in the request object is missing, which
+        // can happen if, for example, the request was created using the Slack modal or web app
+        request.permission.publicKey ?? options.publicKey,
```
GGonryun, line 135: TODO: always rely on local key instead

PR #386, opened 2026-04-13
Update cli to retry polling for request status in case of connection termination

varunalla:
## Problem

When a CLI command requires waiting for a request to complete (e.g. SSH access), the previous implementation used a single streaming connection that combined request submission and status polling into one long-lived HTTP POST. If that connection was interrupted due to a transient network error, the entire flow would fail — requiring the user to re-run the command, which would submit a duplicate request.

## Change

This PR decouples request submission from status polling and adds retry logic to the status polling phase:

- Replaces `fetchStreamingCommand` (a single streaming POST that both created and polled the request) with two distinct steps:
  1. `fetchCommand` — submits the request once, returns a request ID
  2. `fetchStreamingStatus` — polls the request status by ID via a separate GET endpoint
- Adds `retryWithSleep` around `fetchCommand` so transient network errors during submission are also retried
- Adds an outer retry loop around `fetchStreamingStatus` that retries on `"Network error: Unable to reach the server."` for up to 5 minutes, without re-submitting the original request
- Extracts `RETRY_OPTIONS` and `isNetworkError` into shared modules (`constants.ts`, `util.ts`) for reuse across `api.ts` and `request.ts`

Affected areas: **workflow**

## Type of Change

## Validation

Validated manually by:
1. Running a command that waits for request approval (e.g. `p0 request ssh ... --wait`)
2. Once status polling begins, killing the backend service briefly
3. Confirming `"Network error encountered. Retrying..."` is printed
4. Restoring the backend and confirming the CLI resumes polling the same request ID and resolves normally without creating a duplicate request

No duplicate requests are created during retry because `fetchCommand` is called only once, before the retry loop begins.

## Implementation

The two-layer retry design is intentional:

- **Inner layer** (`regenerateWithSleep` in `fetchWithStreaming`): handles low-level TCP errors (`TypeError: terminated`, `fetch failed`) with exponential backoff up to ~121s
- **Outer layer** (`retryWithSleep` wrapping `pollStatus`): kicks in after the inner layer exhausts its retries, retrying the entire poll attempt for up to 5 minutes from first failure

## Tracking

ENG-7707

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@p0security/cli",
-  "version": "0.26.7",
+  "version": "0.26.8",
   "description": "Execute infra CLI commands with P0 grants",
   "main": "index.ts",
   "repository": {
--- a/src/commands/__tests__/request.test.ts
+++ b/src/commands/__tests__/request.test.ts
@@ -8,7 +8,7 @@ This file is part of @p0security/cli
 
 You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
 **/
-import { fetchCommand, fetchStreamingCommand } from "../../drivers/api";
+import { fetchCommand, fetchStreamingStatus } from "../../drivers/api";
 import { print1, print2 } from "../../drivers/stdio";
 import { failure } from "../../testing/yargs";
 import { RequestResponse } from "../../types/request";
@@ -20,7 +20,7 @@ import yargs from "yargs";
 vi.mock("../../drivers/api", async (importOriginal) => ({
   ...(await importOriginal<typeof import("../../drivers/api")>()),
   fetchCommand: vi.fn(),
-  fetchStreamingCommand: vi.fn(), // Add this
+  fetchStreamingStatus: vi.fn(), // Add this
   streamingApiFetch: vi.fn(),
 }));
 vi.mock("../../drivers/auth");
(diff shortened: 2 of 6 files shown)


GGonryun (approved):

dhing1024 (approved):
This certainly still seems a lot better (I approved), but doesn't this not totally solve the problem? If the `executeAndProcessApiRequest` has to retry sending the request, if for example, the API request was received by the backend but the success response was not received by the CLI. Then after a retry, duplicate requests will be created on the backend.

So really the best, most reliable fix is for the backend to handle an incoming request with the same parameters as one it already knows about idempotently and to not create a new one with a different request ID. But if you can do that then it seems like this PR wouldn't actually be needed, correct?

varunalla:
> So really the best, most reliable fix is for the backend to handle an incoming request with the same parameters as one it already knows about idempotently and to not create a new one with a different request ID. But if you can do that then it seems like this PR wouldn't actually be needed, correct?

The chance of backend end creating the permission request and the client not receiving the response with request id would be slim, but not nil. Most scenarios, we are failing to create the permission request when the server is too busy. Then comes the additional overhead we add to each request to debounce using the request parameters(problematic for scripting use cases). this would add additional requirement of storing(caching) users requests for at least 5 mins or so(heuristics change) for implementing the debounce mechanism, which is an overkill at this point, but something we should think i guess.

PR #393, opened 2026-04-19
google: Use select_account login prompt

nbrahms:
**Problem**

Currently, logins via Google Workspace will use whatever selected account exists in the opened browser window.

This can inadvertently lead to login with the incorrect account if the account does not match user intention. This happens, for instance, if the user has a personal profile window logged in to their personal account, rather than their employee account.

**Change**

To prevent this, always force the user to select an account. If the auth is opened in the incorrect window they can follow the OIDC URL in the correct browser window prior to proceeding.

**Type of Change**

**Validation**

Validated via manual login with Google Workspace.

### DIFF
--- a/src/plugins/oidc/login.ts
+++ b/src/plugins/oidc/login.ts
@@ -146,6 +146,7 @@ export const oidcLoginSteps = (
       url: deviceAuthorizationUrl,
     };
   };
+
   const buildOidcTokenRequest = (authorize: AuthorizeResponse) => {
     validateProviderDomain(org);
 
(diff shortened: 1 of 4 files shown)


gergas3 (approved):
LGTM
The behavior of `prompt: "select_account"` is documented here: https://developers.google.com/identity/openid-connect/openid-connect

src/plugins/oidc/login.ts
```diff
@@ -146,6 +146,7 @@ export const oidcLoginSteps = (
       url: deviceAuthorizationUrl,
     };
   };
+
```
GGonryun, line 149: 🤔

GGonryun (approved):

aaronkeisler (approved):

PR #397, opened 2026-04-29
Improvements to rds subcommand logic

dhing1024:
**Problem**

We are preparing to release a new implementation of the postgres integration. However, there are some existing gaps that need to be addressed before we do so. 

**Change**

- Changes the resource key from "pg2" (old internal name for in-progress work) to "postgres" (new permanent resource key)
- Uses `osSafeCommand` for running AWS CLI (generic Windows-compatible wrapper for running terminal commands)
- Adds a `.trim()` to the `PGPASSWORD` output for postgres

**Type of Change**

**Validation**

The RDS subcommands are currently not in use by any known customer, so we don't need to worry too much about backwards compatibility. Tested manually that this works with the new postgres implementation

### DIFF
--- a/src/commands/aws/rds.ts
+++ b/src/commands/aws/rds.ts
@@ -15,7 +15,7 @@ import { parseArn } from "../../plugins/aws/utils";
 import { DbPermissionSpec } from "../../plugins/db/types";
 import { Authn } from "../../types/identity";
 import { PermissionRequest } from "../../types/request";
-import { exec, throwAssertNever } from "../../util";
+import { exec, osSafeCommand, throwAssertNever } from "../../util";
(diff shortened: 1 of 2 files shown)


GGonryun (approved):

src/commands/aws/rds.ts
```diff
@@ -66,7 +66,7 @@ export const rds = (
           y
             .option("arch", {
               type: "string",
-              choices: ["mysql", "pg"] as const,
+              choices: ["mysql", "postgres"] as const,
```
sweeneytr: Tiny nit, move this up and make the type derived from `typeof dbResourceKey[number]`

sweeneytr (approved):

PR #399, opened 2026-04-30
clean up timer blocking the access request from terminating

varunalla:
**Problem**

Making an access request hangs while waiting for approval.
Steps to reproduce:
1. Make an access request
2. Notice that the access requested just hangs and you have the ctrl + c

**Change**

This PR cleans up the dangling timeout operation, used to make sure we are not retrying/waiting for access request creation, from preventing the cli to close.

- [x] workflow
- [ ] lifecycle SDK
- [ ] datastore abstractions

**Type of Change**

- [x] Bug fix: cleans up the timeout operations

**Validation**

Manually testing by creating an access request

**Tracking (optional)**

CUS-264

**Attachments (optional)**

https://github.com/user-attachments/assets/c420913c-f7a1-4d27-a148-dc6764fcde22

### DIFF
--- a/src/commands/shared/request.ts
+++ b/src/commands/shared/request.ts
@@ -215,8 +215,10 @@ export const request =
       throw data;
     };
 
+    const timeoutSleep = sleep(options?.timeOut ?? MAX_REQUEST_TIMEOUT);
+
     const timeoutOperation = async () => {
-      await sleep(options?.timeOut ?? MAX_REQUEST_TIMEOUT);
+      await timeoutSleep;
       throw ACCESS_WAIT_TIMEOUT_ERROR_MESSAGE;
     };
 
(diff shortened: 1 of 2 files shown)


src/commands/shared/request.ts
```diff
@@ -215,8 +215,10 @@ export const request =
       throw data;
     };
 
+    const timeoutSleep = sleep(options?.timeOut ?? MAX_REQUEST_TIMEOUT);
+
     const timeoutOperation = async () => {
-      await sleep(options?.timeOut ?? MAX_REQUEST_TIMEOUT);
+      await timeoutSleep;
```
GGonryun, line 221: does this also automatically kill the sleep after it gets approved? 🤔

GGonryun (approved):

src/commands/shared/request.ts
varunalla: yes. once the request is approved, the code moves into finally block where we kill the sleep.

PR #405, opened 2026-05-15
Use claude mcp add with OAuth flags

ewindisch:
## Summary
- Switch from `claude mcp add-json` to `claude mcp add` with explicit `--client-id`, `--callback-port`, and `--client-secret` OAuth flags — `add-json` doesn't accept oauth fields in its JSON schema (verified against claude 2.1.141)
- Fix spawned process environment: spread `process.env` so `claude` inherits PATH/HOME/NODE_OPTIONS instead of getting a bare `{ MCP_CLIENT_SECRET }` env

## Test plan
- [ ] Run `yarn p0 claude mcp` and verify the generated `~/.claude.json` entry matches the expected `{ type: "http", url, oauth: { clientId, callbackPort } }` shape
- [ ] Verify OAuth flow completes end-to-end (authorize → callback → token exchange)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@p0security/cli",
-  "version": "0.26.11",
+  "version": "0.27.2",
   "description": "Execute infra CLI commands with P0 grants",
   "main": "index.ts",
   "repository": {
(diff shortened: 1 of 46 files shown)


nbrahms (approved):
lgtm

bgberr (approved):

PR #406, opened 2026-05-18
Add `claude` command suite

nbrahms:
**Problem**

P0 is adding support for registering Claude MCP clients with P0, allowing Claude to be used to call MCP servers hosted within P0's agentic gateway.

Currently the P0 controller hosts REST endpoints for client registration and server discovery. However, use of these REST endpoints requires manual cURL.

**Change**

Adds:

`p0 claude mcp add :server -s :scope`

This command:
1. Uses a previously registered MCP client, or registers a new MCP client if none exists on the local user account
2. Installs the referenced server in the local Claude installation, under the appropriate scope

**Type of Change**

- [ ] Documentation update
- [ ] Dependency update
- [ ] Performance improvement
- [ ] Security fix

**Validation**

With a MCP gateway `local` installed, install the built-in P0 MCP server:

`p0 claude mcp add p0-local -s user`

### DIFF
--- a/src/commands/claude/mcp.ts
+++ b/src/commands/claude/mcp.ts
@@ -1,0 +1,168 @@
+/** Copyright © 2024-present P0 Security
+
+This file is part of @p0security/cli
+
+@p0security/cli is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 3 of the License.
+
+@p0security/cli is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
+**/
+import { authFetch, tenantUrl } from "../../drivers/api";
+import { authenticate } from "../../drivers/auth";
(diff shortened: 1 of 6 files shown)


src/commands/claude/mcp.ts
```diff
@@ -0,0 +1,243 @@
...
+
+const getHostname = async () => {
+  const os = getOperatingSystem();
+  switch (os) {
+    case "mac":
+      return (await promisify(exec)("scutil --get LocalHostName")).stdout;
+    case "win":
+      return (await promisify(exec)("hostname")).stdout;
+    case "linux":
+      return (await promisify(exec)("hostname")).stdout;
```
gergas3: nit

```suggestion
    case "win":
    case "linux":
      return (await promisify(exec)("hostname")).stdout;
```

src/commands/claude/mcp.ts
```diff
@@ -0,0 +1,243 @@
...
+type AddMcpServerArgs = yargs.ArgumentsCamelCase<{
+  debug?: boolean;
+  callbackPort: number | undefined;
+  scope: string | undefined;
+  server: string;
+}>;
+
+const CLIENT_PATH = postfixPath("claude/mcp-client.json");
+
+const REDIRECT_PORT = 8080;
```
gergas3: nit: 8080 is often used for development purposes, it could be something more random.

```suggestion
const REDIRECT_PORT = 51215;
```

gergas3 (approved):

PR #407, opened 2026-05-21
p0cli changes to add file-transfer command

Jenn738:
**Problem**

This PR adds the `file-transfer` CLI command to enable a user to upload a file locally to an s3 bucket. This helps with transferring large files faster than using `scp`. 

**Change**

Adds a new `p0 file-transfer <source> <destination>` command.

- The backend approves the access request and returns the destination bucket, bucket region, and a per-user object key alongside an AWS role delegation. The CLI assumes the role locally and uploads to S3 directly.
- Upload uses multipart upload via `@aws-sdk/lib-storage`'s `Upload` with a file stream — supports arbitrarily large files, retries individual parts on transient network failures, and keeps memory flat regardless of file size.
- Generates short-lived presigned GET (5 min) and DELETE (1 hr) URLs alongside the upload — for downstream consumers that don't have AWS credentials of their own.
- Surfaces S3's error response body on upload failure so 403/permission issues show the real reason instead of a bare HTTP status.

**Smaller refinements bundled in:**

- **Step-by-step status output** (`Requesting access...`, `Access approved...`, `Uploading...`) so long waits feel intentional rather than frozen.
- **`httpUploadProgress` listener** prints byte counts and percent during the upload so multi-GB transfers show live progress.
- **Retry-with-backoff on `AccessDenied`** in the upload step. IAM has eventual consistency — a freshly-attached policy can take several seconds to propagate before S3 honors it. The retry covers up to ~30 s with visible logging so the first invocation succeeds transparently.
- Tightened response types: the AWS delegation field is marked optional to match runtime reality, guarded by a runtime check.
- Stream body to the SDK via `createReadStream` so we don't buffer the whole file in memory.
- Sorted union/import order to satisfy lint rules.

- [ ] workflow
- [ ] lifecycle SDK
- [ ] datastore abstractions

**Type of Change**

- [ ] Documentation update
- [x] Dependency update
- [ ] Performance improvement
- [ ] Security fix

**Validation**

Manual end-to-end test against a development environment:

- 100 MB file upload completes via multipart; resulting object appears at the expected key with the correct size.
- 6 GB file upload completes via multipart (the resulting object's `ETag` carries a `-1200` suffix, confirming 1200 stitched parts). Proves we are no longer bounded by S3's 5 GB single-PUT limit.
- Hash check (md5) between the uploaded source and the downloaded copy (via the GET URL) matches.
- The IAM propagation race was observed in practice: the first invocation against a freshly-granted session originally failed with `AccessDenied`. With the retry loop added in this PR, subsequent invocations succeed without manual retry.
- `yarn lint:types` clean. `yarn lint` clean.

**Risks**

- This command relies on new fields (`bucketRegion`, `objectKey`) on the access-request response. The matching server-side change must be deployed before this PR is released. Without it, the CLI will sign requests for an undefined region and S3 will reject them.
- Adds three new `@aws-sdk/*` dependencies. Their versions are aligned (lib-storage has a peer dep on a matching client-s3 minor).

**Tracking (optional)**
https://linear.app/p0-security/issue/CUS-51/splunk-or-aws-file-transfer-initial-cli-logic-for-access-requests
https://linear.app/p0-security/issue/CUS-50/splunk-or-aws-scp-cli-logic-to-upload-file-from-local-to-s3-bucket

**Implementation**

- The upload step does not use a presigned PUT URL even though GET and DELETE remain presigned. The CLI is both the credential-holder and the uploader — packaging the credentials into a single-use URL adds no security boundary, and it forces single-PUT semantics (5 GB cap). Using the SDK directly removes the cap, enables multipart with built-in retries and parallelism, and removes a layer of indirection. GET and DELETE remain presigned because their consumers (the destination process for download, the cleanup step) don't have AWS credentials.
- Bucket region is read from the access-request response rather than probed at upload time. The bucket can live in a different region from the destination instance; the server already knows the bucket's region, so plumbing it through is simpler and faster than a HEAD probe.

**Follow-up Work**

https://linear.app/p0-security/issue/CUS-328/splunk-or-aws-scp-file-transfer-cleanup

### DIFF
--- a/src/commands/file-transfer.ts
+++ b/src/commands/file-transfer.ts
@@ -1,0 +1,97 @@
+/** Copyright © 2024-present P0 Security
+
+This file is part of @p0security/cli
+
+@p0security/cli is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 3 of the License.
+
+@p0security/cli is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
+**/
+import { authenticate } from "../drivers/auth";
+import { print2 } from "../drivers/stdio";
+import { traceSpan } from "../opentelemetry/otel-helpers";
+import {
+  generateTransferUrls,
+  provisionTransferRequest,
+} from "../plugins/file-transfer";
+import * as fs from "fs/promises";
+import yargs from "yargs";
+
+export type FileTransferCommandArgs = {
+  source: string;
+  destination: string;
+  reason?: string;
+  debug?: boolean;
+};
+
+export const fileTransferCommand = (yargs: yargs.Argv) =>
+  yargs.command<FileTransferCommandArgs>(
+    "file-transfer <source> <destination>",
+    "Transfer a local file to a remote instance via a temporary signed-URL bucket.",
+    (yargs) =>
+      yargs
+        .positional("source", {
+          type: "string",
+          demandOption: true,
+          description: "Local file path",
+        })
+        .positional("destination", {
+          type: "string",
+          demandOption: true,
+          description: "Instance ID of the transfer destination",
+        })
+        .option("reason", {
+          type: "string",
+          describe: "Reason access is needed",
+        })
+        .option("debug", {
+          type: "boolean",
+          describe: "Print debug information, including signed URLs.",
+        }),
+    fileTransferAction
+  );
+
+const fileTransferAction = async (
+  args: yargs.ArgumentsCamelCase<FileTransferCommandArgs>
+) => {
+  await traceSpan(
+    "file-transfer.command",
+    async (span) => {
+      span.setAttribute("source", args.source);
+      span.setAttribute("destination", args.destination);
+
+      const authn = await authenticate(args);
+
+      const target = await provisionTransferRequest(authn, args);
+
+      const { putUrl, getUrl, deleteUrl } = await generateTransferUrls(
+        authn,
+        target,
+        args.debug
+      );
+
+      if (args.debug) {
+        print2(`PUT    (5m): ${putUrl}`);
+        print2(`GET    (5m): ${getUrl}`);
+        print2(`DELETE (1h): ${deleteUrl}`);
+      }
+
+      const buffer = await fs.readFile(args.source);
+      const uploadResponse = await fetch(putUrl, {
+        method: "PUT",
+        body: buffer as unknown as BodyInit,
+      });
+
+      if (!uploadResponse.ok) {
+        const body = await uploadResponse.text().catch(() => "");
+        throw `Upload failed: ${uploadResponse.status} ${uploadResponse.statusText}${body ? `\n${body}` : ""}`;
+      }
+
+      print2("Uploaded.");
+    },
+    {
+      command: "file-transfer",
+    }
+  );
+};
--- a/src/commands/file-transfer.ts
+++ b/src/commands/file-transfer.ts
@@ -18,6 +18,7 @@ import {
 } from "../plugins/file-transfer";
 import { Upload } from "@aws-sdk/lib-storage";
 import { createReadStream, statSync } from "fs";
+import { basename } from "node:path";
 import yargs from "yargs";
 
 export type FileTransferCommandArgs = {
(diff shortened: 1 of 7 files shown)


src/commands/file-transfer.ts
```diff
@@ -0,0 +1,157 @@
...
+
+      print2("Requesting file-transfer access...");
+      const target = await provisionTransferRequest(authn, args);
+      print2(`Access approved for s3://${target.bucket}/${target.key}`);
+
+      print2("Preparing upload credentials...");
+      const { s3, getUrl, deleteUrl, expirySeconds } =
+        await generateTransferUrls(authn, target, args.debug);
+
+      const fmt = (s: number) =>
```
melissarh57: could we rename function to a more descriptive name?
```suggestion
      const renderDuration = (s: number) =>
```

src/commands/file-transfer.ts
```diff
@@ -0,0 +1,157 @@
...
+        print2(`DELETE (${fmt(expirySeconds.delete)}): ${deleteUrl}`);
+      }
+
+      print2(`Uploading ${args.source}...`);
+
+      // The backend grants the AWS role permission to write to our prefix, but
+      // IAM has eventual consistency — the policy can take several seconds to
+      // propagate before S3 honors it. Retry AccessDenied for up to 30s so the
+      // first invocation just works instead of failing the user.
+      const GRANT_PROPAGATION_TIMEOUT_MS = 30_000;
```
melissarh57: can we use the common/retry function here?

melissarh57: maybe retryWithSleep

src/plugins/file-transfer/types.ts
```diff
@@ -0,0 +1,52 @@
...
+};
+
+export type TransferTarget = {
+  bucket: string;
+  key: string;
+  region: string;
+  awsSpec: AwsResourcePermissionSpec;
+};
+
+export type TransferUrls = {
```
melissarh57: i would propose getting rid of this type or renaming it. b/c it's not just urls but the whole s3 client, but maybe since the type is just used in one place we can get rid of it?

src/commands/file-transfer.ts
```diff
@@ -0,0 +1,157 @@
...
+
+      print2("Preparing upload credentials...");
+      const { s3, getUrl, deleteUrl, expirySeconds } =
+        await generateTransferUrls(authn, target, args.debug);
+
+      const fmt = (s: number) =>
+        s >= 3600 ? `${Math.round(s / 3600)}h` : `${Math.round(s / 60)}m`;
+      if (args.debug) {
+        print2(`GET    (${fmt(expirySeconds.get)}): ${getUrl}`);
+        print2(`DELETE (${fmt(expirySeconds.delete)}): ${deleteUrl}`);
```
melissarh57: i forget if u said we'll get rid of this or not but making a note here so we don't forget. We won't want to log these in debug since they r creds once we roll this feature out. could be worth TODO or something so we don't accidentally since it wouldn't look good to customers

melissarh57 (approved):
looks good! but yes let's have someone else review as well for posterity. 

nice job!

src/plugins/file-transfer/index.ts
```diff
@@ -0,0 +1,111 @@
...
+    { message: "approval-required" }
+  );
+
+  if (!response) {
+    throw "Did not receive a response from server";
+  }
+
+  const awsSpec = response.request.delegation.aws;
+  if (!awsSpec) {
+    throw "Backend granted file-transfer access, but did not provide AWS delegation";
```
MichaelDimitras: I don't think we should use the term `delegation` here if this the message the end user will see.  I think something more explaining that there was an error getting access in AWS would be better.

src/commands/file-transfer.ts
Jenn738: going to go with `renderDurationSec`

src/commands/file-transfer.ts
Jenn738: yep!

src/plugins/file-transfer/types.ts
```diff
@@ -0,0 +1,52 @@
...
+    aws?: AwsResourcePermissionSpec;
+  };
+};
+
+export type TransferTarget = {
+  bucket: string;
+  key: string;
+  region: string;
+  awsSpec: AwsResourcePermissionSpec;
+};
```
MichaelDimitras: Like Melissa's comment below, I think it might be cleaner to remove this type and just inline it since it's only used in one place.

If `generateTransferUrls` is updated with the inline type, then the return of `provisionTransferRequest` doesn't need to be typed at all.

Definitely not a must do, but seems more consistent with our pattern of only creating abstractions when used in two places.

```
export const generateTransferUrls = async (
  authn: Authn,
  target: {
    bucket: string;
    key: string;
    region: string;
    awsSpec: AwsResourcePermissionSpec;
  },
```

MichaelDimitras (approved):

src/commands/file-transfer.ts
Jenn738: I've noted it down as a small task to unblock deployment: https://linear.app/p0-security/issue/CUS-328/splunk-or-aws-scp-file-transfer-cleanup

src/plugins/file-transfer/index.ts
melissarh57: agreed!

melissarh57 (approved):

melissarh57 (approved):

PR #408, opened 2026-05-27
add support for multiple delegates

GGonryun:
**Problem**

The main app is changing its delegation data model: `request.delegation` will start arriving as an array of `{ key, request }` entries rather than as a record keyed by integration name. Without this change, the CLI's `kubeconfig`, `aws rds generate-db-auth-token`, and AWS SSH flows would silently break the moment the backend rolls out the new shape — `request.delegation.aws` and `access.delegation["aws-rds"]` would return `undefined`, and users would see "not an EKS cluster" / "not a RDS instance" errors despite a valid grant.

**Change**

Introduces a small abstraction over the delegation field so both shapes work during (and after) the backend migration:

- New `src/types/delegation.ts` exports `DelegationField<Old>` (a union of the legacy record form and the new array form) and a `getDelegate(delegation, key)` helper that resolves a delegate by key from either shape.
- `PermissionSpec.delegation` is now `DelegationField<D>` and optional, with `D`'s constraint widened to allow `undefined`-valued entries (lets `K8sPermissionSpec` declare `{ aws?: AwsResourcePermissionSpec }`).
- `K8sPermissionSpec`, `DbPermissionSpec`, and `AwsSshPermissionSpec` updated to accept either shape.
- The four call sites (`commands/kubeconfig.ts`, `commands/aws/rds.ts`, `plugins/aws/ssh.ts`) switched to `getDelegate`. The AWS SSH path also gains an explicit throw when `accountId` is missing — previously implicit, surfaced by the helper's `| undefined` return.

**Type of Change**

- [ ] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] Dependency update
- [ ] Performance improvement
- [ ] Security fix

**Validation**

Automated tests:

- `getDelegate` unit tests covering record form, array form, nested + mixed nesting, malformed entries, nullish input (12 tests).
- Type-level assertions confirming both shapes assign to `K8sPermissionSpec`, `DbPermissionSpec`, `AwsSshPermissionSpec` (8 tests — compile-time check).
- Pure-function tests for `awsSshProvider.requestToSsh` covering both shapes, the `resource` fallback path, and the missing-`accountId` error path (7 tests).
- yargs-driven integration tests for `kubeconfigAction` and `rds generate-db-auth-token` that mock the network/exec/file boundaries and assert `awsCloudAuth` receives the same `AwsResourcePermissionSpec` regardless of which shape the backend sent. RDS covers all four nesting combinations (record/record, array/array, array/record, record/array) — 9 tests.

All 158 tests in the suite pass. `tsc --noEmit` is clean.

Risk: low. The change is additive — the helper accepts either shape, and every consumer that reads delegation already routes through it. Construction sites in tests and any code that builds permission specs still work because assigning a required field to an optional slot is always allowed.

Tested with actual commands

```
./p0 kubeconfig --cluster dev-test-cluster-1 \
 --resource '* / * / *' \
 --role 'ClusterRole / admin' \
 --reason 'miguel said so' \
 --duration '8 hours'

./p0 ssh random-id-goes-here

./p0 aws rds generate-db-auth-token --arch="postgres" --role="pg_read_all_data" --database="postgres" --instance="p0-test-1-aurora-pg"
```

**Tracking (optional)**

https://linear.app/p0-security/issue/CUS-337/cli-support-multiple-delegates

**Implementation (optional)**

- `getDelegate`'s generics are factored as `<K extends string, V>` (with `K` pinned to the key argument and `V` inferred from the value) rather than `<Old, K extends keyof Old>`. The latter caused TS 4.8 to bidirectionally infer `Old` as the array branch when matching the union `DelegationField<Old>`, breaking the key constraint. The current shape avoids that.
- `DbPermissionSpec` still spells out its fields directly rather than extending `PermissionSpec` — the inner `aws-rds` delegate doesn't carry `type` or `generated` in the current type, so wrapping it in `PermissionSpec` would require confirming the backend actually sends those fields.

**Follow-up Work (optional)**

- Once the backend has fully cut over to the array form, the record-form branch of `DelegationField` can be removed.
- Revisit `DbPermissionSpec`/`AwsRdsDelegate` to align with `PermissionSpec` if the backend confirms `type`/`generated` are present on the nested `aws-rds` delegate.

### DIFF
--- a/src/types/delegation.ts
+++ b/src/types/delegation.ts
@@ -23,11 +23,11 @@ export type DelegationEntry<K extends string, R> = {
 /** Delegation field that tolerates both the legacy record form and the new
  * array form. Callers should not read this directly — use {@link getDelegate}.
  */
-export type DelegationField<Old extends Record<string, any>> =
+export type DelegationField<Spec extends Record<string, any>> =
   | {
-      [K in keyof Old & string]: DelegationEntry<K, Old[K]>;
-    }[keyof Old & string][]
-  | Old;
+      [K in keyof Spec & string]: DelegationEntry<K, Spec[K]>;
+    }[keyof Spec & string][]
+  | Spec;
 
 /** Resolve a delegate by key, accepting either the legacy record-form
  * delegation or the new array-form delegation.
@@ -35,9 +35,9 @@ export type DelegationField<Old extends Record<string, any>> =
  * Returns the underlying delegate value (with `permission`, `generated`,
  * and nested `delegation` fields), or `undefined` if no entry matches.
  *
- * The generic shape (`K`, `V` rather than the full `Old` record) is
- * deliberate: matching the union `DelegationField<Old>` bidirectionally
- * confuses TS's inference and can lock `Old` onto the array branch.
+ * The generic shape (`K`, `V` rather than the full `Spec` record) is
+ * deliberate: matching the union `DelegationField<Spec>` bidirectionally
+ * confuses TS's inference and can lock `Spec` onto the array branch.
  * Pinning `K` to the key argument and inferring `V` from the value avoids
  * that.
  */
--- a/src/types/delegation.ts
+++ b/src/types/delegation.ts
@@ -42,8 +42,7 @@ export type DelegationField<Old extends Record<string, any>> =
  * that.
  */
 export const getDelegate = <K extends string, V>(
-  delegation:
-    DelegationEntry<K, V>[] | { [P in K]?: V } | null | undefined,
+  delegation: DelegationEntry<K, V>[] | { [P in K]?: V } | null | undefined,
   key: K
 ): V | undefined => {
   if (delegation == null) return undefined;
--- a/src/types/delegation.ts
+++ b/src/types/delegation.ts
@@ -1,0 +1,55 @@
+/** Copyright © 2024-present P0 Security
+
+This file is part of @p0security/cli
+
+@p0security/cli is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 3 of the License.
+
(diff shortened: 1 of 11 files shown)


melissarh57 (approved):
thank youuuu

src/types/delegation.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+export type DelegationEntry<K extends string, R> = {
+  key: K;
+  request: R;
+};
+
+/** Delegation field that tolerates both the legacy record form and the new
+ * array form. Callers should not read this directly — use {@link getDelegate}.
+ */
+export type DelegationField<Old extends Record<string, any>> =
```
nbrahms:
```suggestion
export type DelegationField<Spec extends Record<string, any>> =
```

we probably don't want to use `Old` once we remove the legacy data model

src/types/delegation.ts
```diff
@@ -0,0 +1,54 @@
...
+ * Pinning `K` to the key argument and inferring `V` from the value avoids
+ * that.
+ */
+export const getDelegate = <K extends string, V>(
+  delegation: DelegationEntry<K, V>[] | { [P in K]?: V } | null | undefined,
+  key: K
+): V | undefined => {
+  if (delegation == null) return undefined;
+  if (Array.isArray(delegation)) {
+    const entry = delegation.find((e) => e?.key === key);
```
nbrahms, line 50: What happens if two match?

GGonryun: .find always returns the first thing we match, and that is expected here as we're supporting these older use cases.

ideally none of this code would be in our CLI and our backend api layer can be responsible for making this determination. i don't like that our backend data abstractions are leaking into the different clients we support.

PR #409, opened 2026-05-27
Gus boling/refresh okta token

gusboling:
**Problem**

Users authenticating with the CLI using Okta may be logged out before their session expires because the token we receive from Okta has a shorter TTL than the session itself. This results in, among other things, users being forcibly logged out of long-running SSH sessions. Since we don't want to unnecessarily interrupt workflows while the authenticating Okta session is still active, we should attempt to refresh tokens in the background if possible.

**Change**

When requesting a new OIDC token from Okta, the CLI will now attempt to request the `offline_access` scope in addition to existing scopes. This should provide us with a refresh-token that will allow us to retrieve a fresh token once the existing one expires, assuming that the underlying Okta session is still active. If we are unable to request this new scope, or are unable to refresh the token for any reason (including the session ending), we will default to the existing behavior, and the user will have to re-authenticate manually.

Additionally, when the user runs `p0 logout` we will make a best-effort attempt to revoke the token provided by Okta.

- [X] authentication / authorization
- [X] workflow
- [ ] lifecycle SDK
- [ ] datastore abstractions

**Type of Change**

- [X] Bug fix (non-breaking change that fixes an issue)
- [X] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] Dependency update
- [ ] Performance improvement
- [ ] Security fix

**Validation**

With the new code we've updated unit tests and added new ones.

**Tracking (optional)**

https://linear.app/p0-security/issue/CUS-256

**Implementation (optional)**

- This PR only covers refreshing tokens provided by Okta (see ticket for context). This PR also doesn't impact authentication for the P0 web UI, and is focused specifically on CLI use cases.
- Since we will be periodically updating the `identity.json` file with a fresh token, the file operation involved has been revised to be more atomic and less susceptible to a race condition existing across multiple CLI processes.
- Since there are some edge cases (token refresh failures, etc.) that may be hard to reproduce, additional debug logging has been added to preemptively aid with troubleshooting.


src/drivers/auth/index.ts
```diff
@@ -157,14 +191,31 @@ export const writeIdentity = async (
...
+  // local copy.
+  try {
+    const identity = await loadCredentials();
+    if (
+      identity.credential.refresh_token &&
+      getProviderType(identity.org) === "okta"
+    ) {
+      await revokeOktaRefreshToken(identity, { debug: options?.debug });
+    }
+  } catch {
```
GGonryun: since we have debug here, should we log the error if the flag is enabled?

src/commands/logout.ts
```diff
@@ -39,15 +36,11 @@ const safeDelete = async (
 const logout = async (debug: boolean): Promise<void> => {
   print2("Logging out...");
 
-  const identityPath = getIdentityFilePath();
-  await safeDelete(identityPath, "identity file", debug);
```
GGonryun, line 43: i think we also need to delete the files, the logout does a full clean up to prevent any extra files from being interpreted as "logged in" and to help troubleshoot caching related issues.

src/drivers/auth/refresh.ts
```diff
@@ -0,0 +1,173 @@
...
+  } catch (e) {
+    throw refreshError("network_error", { cause: e });
+  }
+
+  if (!response.ok) {
+    let detail: string | undefined;
+    try {
+      detail = await response.text();
+    } catch {
+      // ignore
```
GGonryun: since we pass down debug we can probably still log something here if there's an error, same with the other catch blocks. would be useful to be verbose in case something goes wrong.

GGonryun (approved):

src/commands/logout.ts
gusboling: That happens as part of deleteIdentity, but I think we can refactor this to make it more clear.

src/drivers/auth/index.ts
gusboling: Looked into it - this particular try/catch is superfluous, since we're already catching all of our errors in the `revokeOktaRefreshToken` function and not throwing any new ones.

We also log any issues in that function, so I think we can remove this try/catch entirely.

src/drivers/auth/refresh.ts
gusboling: Fixed; I've consolidated the logic a bit, since we're really only hedging against a response being returned without any details about why it failed.

I've converted it to a one-liner for readability, but I think we're already logging everything we can/should on line 101

PR #410, opened 2026-06-01
[CUS-54] Added file-transfer cli logic to invoke the download command to ec2 instance

Jenn738:
**Problem**

The `file-transfer` command previously only handled half of the workflow: it uploaded a local file to a scoped S3 bucket and printed signed GET/DELETE URLs for the user to use manually. This change extends file-transfer to actually land the file on a target instance using a GET presigned url and `curl`.

**Change**

Extends `file-transfer` to complete the round trip in a single command:

1. Upload the local file to S3 (unchanged).
2. Request SSH access to the destination instance via the existing SSH approval flow.
3. Once approved, sign a fresh GET URL and have the instance pull the file via `curl` to the user's home directory.

The GET URL is now signed **after** the SSH approval clears rather than up front, so its 5-minute TTL covers the download window instead of being burned during the approval wait.

Also updates `delegation.aws` access to go through the typed `getDelegate` helper, matching how other plugins read delegation specs.

- [x] authentication / authorization (uses SSH approval flow for the download leg)
- [x] workflow (file-transfer now composes upload + SSH download)
- [ ] lifecycle SDK
- [ ] datastore abstractions

**Type of Change**

- [ ] Documentation update
- [ ] Dependency update
- [ ] Performance improvement
- [ ] Security fix

**Validation**

Manual end-to-end test against a dev environment:
- Uploaded a file with `file-transfer <source> <destination>`, confirmed S3 upload succeeded.
- Confirmed SSH approval prompt appeared for the destination instance.
- After approval, confirmed file landed at `/home/<linuxUser>/<basename>` on the target.
- Confirmed the command exits cleanly (no hangs from orphaned `session-manager-plugin` processes).

Error paths covered:
- Missing `curl` on the remote (exit code 127) surfaces a clear error pointing at the followup ticket for a fallback downloader.
- Non-zero remote exit codes propagate with the code in the error message.

Risks:
- The download command is hardcoded to `curl` for now. Most mainstream EC2 AMIs ship with curl, but air-gapped or minimal images may not. Tracked as followup.
- The forced `exitProcess(0)` mirrors what the SSH commands already do for the same orphaned-child-process reason.

**Tracking (optional)**

**Implementation (optional)**

- Splits signed-URL generation: `generateSignedUrl` called to only sign the DELETE URL up front; `signGetUrl` is called later, after SSH approval, so the GET TTL isn't consumed by approval wait.
- The SSH leg reuses `prepareRequest` + `sshOrScp` from the existing SSH plumbing. `source` is stripped from args before reuse because `createCommand` routes on `"source" in args` to pick scp vs ssh, and we want the ssh branch here.
- `FileTransferPermissionSpec` moves `delegation.aws` from an intersection type into the `PermissionSpec` generic's delegation slot, matching the shape other permission specs use.

### DIFF
--- a/src/plugins/file-transfer/index.ts
+++ b/src/plugins/file-transfer/index.ts
@@ -10,6 +10,7 @@ You should have received a copy of the GNU General Public License along with @p0
 **/
 import { FileTransferCommandArgs } from "../../commands/file-transfer";
 import { request } from "../../commands/shared/request";
+import { getDelegate } from "../../types/delegation";
 import { Authn } from "../../types/identity";
 import { PermissionRequest } from "../../types/request";
 import { awsCloudAuth } from "../aws/auth";
@@ -52,7 +53,7 @@ export const provisionTransferRequest = async (
     throw "Did not receive a response from server";
   }
 
-  const awsSpec = response.request.delegation.aws;
+  const awsSpec = getDelegate(response.request.delegation, "aws");
   if (!awsSpec) {
     throw "Backend granted file-transfer access, but there was an error getting AWS access details";
   }
--- a/src/plugins/file-transfer/index.ts
+++ b/src/plugins/file-transfer/index.ts
@@ -79,9 +79,8 @@ export const generateTransferUrls = async (
   debug?: boolean
 ): Promise<{
   s3: S3Client;
-  getUrl: string;
   deleteUrl: string;
-  expirySeconds: { get: number; delete: number };
+  expirySeconds: { delete: number };
 }> => {
   const credentials = await awsCloudAuth(authn, target.awsSpec, debug);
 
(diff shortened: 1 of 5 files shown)


src/plugins/file-transfer/index.ts
```diff
@@ -127,29 +128,23 @@ export const generateTransferUrls = async (
         `Check your system clock or re-run the request.`
     );
   }
-  const secondsToExpireGetUrl = Math.min(SECONDS_TO_EXPIRE_GET_URL, remaining);
-  const secondsToExpireDeleteUrl = Math.min(
-    SECONDS_TO_EXPIRE_DELETE_URL,
-    remaining
-  );
+  const maxExpiry =
```
melissarh57, line 131: changed so function only generates either get or delete url not both since need at differetn times

PR #412, opened 2026-06-02
add support for different shells like fish

GGonryun:
**Problem**

Users whose login shell is **fish** can't populate AWS credentials from the CLI's credential-export commands. `p0 aws role assume` (and `permission-set` / `rds`) print POSIX/bash syntax — `export KEY=value` and `eval $(...)` — which fish cannot parse, so the variables silently never get set and downstream AWS calls fall back to the default credential chain (often a 403). Investigation confirmed the `p0 ssh` transport itself is **not** shell-dependent (credentials are injected directly into the child process env), so this is scoped to the printed credential/instruction snippets users copy or `eval`.

**Change**

Adds a small shell-aware formatter (`detectShell` + `newShellFormatter` in `util.ts`) and routes the three credential/instruction printers through it:
- `printAwsCredentials` (`aws role assume`, `aws permission-set`)
- `awsSshProvider.reproCommands` (debug repro hint)
- `aws rds` psql/mysql connection instructions

fish users now get `set -gx KEY value` and `cmd | source`; everything else (bash/zsh/sh, unset `$SHELL`) keeps the existing POSIX output. Detection is fail-safe — only an exact `fish` match opts out of the legacy POSIX path.

**Type of Change**

**Validation**

Automated unit tests added/extended:
- `util.test.ts` — `detectShell` (posix shells, exact-fish match, no false positives, unset) and the posix/fish formatters (posix output pinned byte-for-byte to the legacy strings).
- `role.test.ts` — bash + fish snapshots for `printAwsCredentials`.
- `aws/ssh.test.ts` — `reproCommands` for bash, fish, and IDC.
- `rds.test.ts` — shell-aware instructions for bash/postgres and fish postgres/mysql.

`tsc`, ESLint, and Prettier are clean. Risk to non-fish users is minimal: detection defaults to POSIX for everything but a literal fish login shell, the`export KEY=value` lines and `${VAR}` references are unchanged, and none of this code is executed by the CLI — these are strings printed for the user.

**Tracking**

[CUS-359](https://linear.app/p0-security/issue/CUS-359/p0-cli-expects-bash-shell)

**Implementation**

Formatter is modeled as an interface + plain object literals (one per shell) selected via a `Record<ShellKind, ShellFormatter>` map, rather than a class or inline switches.

### DIFF
--- a/src/__tests__/util.test.ts
+++ b/src/__tests__/util.test.ts
@@ -14,19 +14,17 @@ import { describe, expect, it } from "vitest";
 const asEnv = (env: Record<string, string>) => env as NodeJS.ProcessEnv;
 
 describe("detectShell", () => {
-  it("treats every common POSIX shell as posix", () => {
-    for (const shell of [
-      "/bin/bash",
-      "/usr/bin/bash",
-      "/usr/bin/zsh",
-      "/bin/zsh",
-      "/bin/sh",
-      "/bin/dash",
-      "/usr/bin/ksh",
-      "/usr/bin/bash", // git-bash on Windows reports this
-    ]) {
-      expect(detectShell(asEnv({ SHELL: shell }))).toBe("posix");
-    }
+  it.each([
+    "/bin/bash",
+    "/usr/bin/bash",
+    "/usr/bin/zsh",
+    "/bin/zsh",
+    "/bin/sh",
+    "/bin/dash",
+    "/usr/bin/ksh",
+    "/usr/bin/bash", // git-bash on Windows reports this
+  ])("treats %s as posix", (shell) => {
+    expect(detectShell(asEnv({ SHELL: shell }))).toBe("posix");
   });
 
   it("detects fish only on an exact basename match (no false positives)", () => {
--- a/src/__tests__/util.test.ts
+++ b/src/__tests__/util.test.ts
@@ -44,7 +44,7 @@ describe("detectShell", () => {
   });
 });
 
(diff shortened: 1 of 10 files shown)


src/__tests__/util.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
+**/
+import { detectShell, newShellFormatter } from "../util";
+import { describe, expect, it } from "vitest";
+
+const asEnv = (env: Record<string, string>) => env as NodeJS.ProcessEnv;
+
+describe("detectShell", () => {
+  it("treats every common POSIX shell as posix", () => {
```
melissarh57: optional nit: could use it.each for clearer debugging which failed


```suggestion
it.each([
  "/bin/bash",
  "/usr/bin/bash",
  "/usr/bin/zsh",
  "/bin/zsh",
  "/bin/sh",
  "/bin/dash",
  "/usr/bin/ksh",
])("treats %s as posix", (shell) => {
  expect(detectShell(asEnv({ SHELL: shell }))).toBe("posix");
});
```

melissarh57: oops this suggestion didn't replace correctly don't apply

melissarh57 (approved):
Great! Just the one questions about installers otherwise looks good nice test coverage :)

PR #413, opened 2026-06-04
automate google cloud auth login

GGonryun:
**Problem**

Connecting to a Google Cloud (GCP) instance with `p0 ssh` requires valid Google Cloud
CLI credentials — the CLI shells out to `gcloud` to import the OS Login SSH key and to
open the IAP tunnel. Logging in to P0 itself (via SSO such as Okta, or any other provider)
does **not** establish those credentials: they come from a separate `gcloud auth login`.

Previously the CLI assumed the user had already run `gcloud auth login` and, if they
hadn't, simply failed partway through with a hint to go run it manually. AWS and Azure
already perform their cloud-provider login automatically during `p0 ssh`; GCP did not.

**Change**

`p0 ssh` to a GCP target now triggers `gcloud auth login` automatically when the gcloud
credentials are missing/expired, and skips it when they are still valid — matching the
AWS/Azure behavior.

- New `ensureGcloudLogin` helper: checks validity with `gcloud auth print-access-token`
  (which succeeds whenever gcloud has valid or silently-refreshable credentials) and runs
  the interactive `gcloud auth login` only when that check fails. It returns a valid access
  token so callers don't fetch it twice.
- Wired in at both points a gcloud token is needed: the OS Login key import (covers the
  direct `p0 ssh` and `ssh-resolve` flows, which run before the cloud-login hook) and the
  GCP provider's `cloudProviderLogin` hook (covers the `ssh-proxy` ProxyCommand flow).
- The interactive login routes the child's stdout to stderr so it can never corrupt the
  SSH data channel that `ssh-proxy` carries on stdout.

- [ ] workflow
- [ ] lifecycle SDK
- [ ] datastore abstractions

**Type of Change**

- [ ] Documentation update
- [ ] Dependency update
- [ ] Performance improvement
- [ ] Security fix

**Validation**

- Added unit tests (`src/plugins/google/__tests__/auth.test.ts`, 5 cases) covering: skip
  when credentials are valid; run login when invalid and return a fresh token; the token
  check never prints the token (debug forced off); and helpful rejections when login exits
  non-zero or fails to spawn. The new tests pass.
- `tsc` and `eslint --max-warnings 0` are clean on the changed files.
- Manual end-to-end against a live GCP instance was **not** performed in this session;
  reviewers connecting to a real GCP target should verify the login triggers after
  `gcloud auth revoke --all` and is skipped when already authenticated.
- Low blast radius: changes are confined to the GCP SSH provider plus a one-line `debug`
  pass-through in the shared `sshProxy`; other providers (AWS/Azure/self-hosted) are
  untouched. The pre-existing `gcloud auth login` hint paths are kept as a fallback.
- Also verified that the login prompt doesn't appear for non-gcloud requests

**Tracking (optional)**

[CUS-382](https://linear.app/p0-security/issue/CUS-382/p0-cli-automate-google-cloud-auth-login)

**Implementation (optional)**

- The gcloud token is first needed during OS Login key import, which runs *before* the
  generic `cloudProviderLogin` hook — and `ssh-resolve` never calls that hook while
  `ssh-proxy` never imports the key. Hence the helper is invoked at both sites; it's
  idempotent and cheap (a single token check) when already logged in.
- `ensureGcloudLogin` returns the access token so the key-import path makes one call
  instead of an ensure-then-fetch double round-trip.

**Follow-up Work (optional)**

- OS Login imports the key for whatever account `gcloud config get-value account` returns;
  `ensureGcloudLogin` guarantees *a* valid gcloud login but does not verify it matches the
  identity the access was approved for. Validating/auto-selecting the account is out of
  scope here and a candidate for a follow-up.

### DIFF
--- a/src/plugins/google/auth.ts
+++ b/src/plugins/google/auth.ts
@@ -24,7 +24,13 @@ const runGcloudLogin = async ({ debug }: { debug?: boolean }) =>
     print2("Logging in to Google Cloud CLI...");
     const { command, args } = gcloudCommandArgs(["auth", "login"]);
     const child = spawnWithCleanEnv(command, args, {
-      // [stdin, stdout, stderr]: route child stdout to OUR stderr (never fd 1)
+      // stdio is [stdin, stdout, stderr]. We send the child's stdout to OUR
+      // stderr instead of inheriting fd 1: `gcloud auth login` writes its
+      // human-readable progress to stdout, but this CLI reserves fd 1 for
+      // machine-readable output (e.g. access tokens, JSON) that callers parse.
+      // Inheriting the child's stdout would interleave gcloud's chatter into
+      // that stream and corrupt it, so we redirect it to stderr — where
+      // human-facing text belongs.
       stdio: ["inherit", process.stderr, "inherit"],
     });
     child.on("error", (error) =>
--- a/src/plugins/google/auth.ts
+++ b/src/plugins/google/auth.ts
@@ -1,0 +1,60 @@
+/** Copyright © 2024-present P0 Security
+
+This file is part of @p0security/cli
+
+@p0security/cli is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 3 of the License.
+
+@p0security/cli is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
+**/
+import { asyncSpawn } from "../../common/subprocess";
+import { print2 } from "../../drivers/stdio";
+import { spawnWithCleanEnv } from "../../util";
+import { gcloudCommandArgs } from "./util";
+
+export const getGcloudAccessToken = async (): Promise<string> => {
+  const { command, args } = gcloudCommandArgs(["auth", "print-access-token"]);
+  // Force debug=false otherwise it prints the access token
+  return await asyncSpawn({ debug: false }, command, args);
+};
+
+const runGcloudLogin = async ({ debug }: { debug?: boolean }) =>
+  new Promise<void>((resolve, reject) => {
+    print2("Logging in to Google Cloud CLI...");
+    const { command, args } = gcloudCommandArgs(["auth", "login"]);
+    const child = spawnWithCleanEnv(command, args, {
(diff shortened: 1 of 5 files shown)


src/plugins/ssh/index.ts
```diff
@@ -696,7 +696,7 @@ export const sshProxy = async (args: {
   const { authn, sshProvider, request, requestId, debug } = args;
 
   const credential: AwsCredentials | undefined =
-    await sshProvider.cloudProviderLogin(authn, request);
+    await sshProvider.cloudProviderLogin(authn, request, debug);
```
Jenn738, line 699: is adding debug here mostly something missing that you caught?

src/plugins/google/auth.ts
```diff
@@ -0,0 +1,60 @@
...
+**/
+import { asyncSpawn } from "../../common/subprocess";
+import { print2 } from "../../drivers/stdio";
+import { spawnWithCleanEnv } from "../../util";
+import { gcloudCommandArgs } from "./util";
+
+export const getGcloudAccessToken = async (): Promise<string> => {
+  const { command, args } = gcloudCommandArgs(["auth", "print-access-token"]);
+  // Force debug=false otherwise it prints the access token
+  return await asyncSpawn({ debug: false }, command, args);
```
Jenn738, line 19: does this mean that whenever `asyncSpawn` is called anywhere else, the user would need to force `debug=false`? Or is this just for `gcloudCommandArgs`? I'm wondering if we'll potentially be relying on a developer in the future to know to set this explicitly.

GGonryun: this is just for cloud command args, this is why i wrapped it in a function, ideally people shouldn't be trying to call gcloud print-access-token without using this command

src/plugins/ssh/index.ts
GGonryun: it's so that we can log if we have to login or skip login, it'll help troubleshoot situations where we might have unexpected behavior.

src/plugins/google/auth.ts
```diff
@@ -0,0 +1,60 @@
...
+};
+
+const runGcloudLogin = async ({ debug }: { debug?: boolean }) =>
+  new Promise<void>((resolve, reject) => {
+    print2("Logging in to Google Cloud CLI...");
+    const { command, args } = gcloudCommandArgs(["auth", "login"]);
+    const child = spawnWithCleanEnv(command, args, {
+      // [stdin, stdout, stderr]: route child stdout to OUR stderr (never fd 1)
```
melissarh57: can u update the comment to give ssh proxy context as to why we need to do this so clearer to readers?

melissarh57 (approved):
all manual testing scenarios passed, code looks good!
just the one comment to improve the comment "why"

PR #415, opened 2026-06-04
Propagate AWS credential expiry and use it to harden file transfers

Jenn738:
**Problem**
Temporary AWS credentials carry an expiry that the CLI wasn't tracking. Surfacing it lets the file-transfer command (a) refresh credentials during uploads that run longer than a single credential's lifetime, and (b) ensure signed download/cleanup URLs never outlive the credentials that signed them

**Change**
1. Plumb credential expiry through 
Okta SAML path now reads STS Expiration (ISO string → epoch ms). The IDC path already returned epoch ms, so both paths now agree.
ssh spawn strips expiresAt before spreading credentials into the child process env (it's metadata, not an env var).

2. Auto-refreshing S3 client (createTransferClient)
The S3 client now takes a credentials provider function that returns expiration. The AWS SDK uses that to re-fetch fresh credentials mid-upload instead of failing — fixes long uploads.

3. Sign download/cleanup URLs after upload, capped to credential lifetime

Tests
Updated the assumeRole mock + tests to cover expiresAt (asserts the ISO Expiration is converted to epoch ms).
Added file-transfer plugin tests for the new client + URL-signing behavior.

### DIFF
--- a/src/commands/file-transfer.ts
+++ b/src/commands/file-transfer.ts
@@ -13,6 +13,7 @@ import { authenticate } from "../drivers/auth";
 import { print2 } from "../drivers/stdio";
 import { traceSpan } from "../opentelemetry/otel-helpers";
 import {
+  createTransferClient,
   generateTransferUrls,
   provisionTransferRequest,
 } from "../plugins/file-transfer";
@@ -88,22 +89,7 @@ const fileTransferAction = async (
       const uploadKey = `${target.prefix}${basename(args.source)}`;
 
       print2("Preparing upload credentials...");
-      const { s3, getUrl, deleteUrl, expirySeconds } =
-        await generateTransferUrls(
-          authn,
-          { ...target, key: uploadKey },
-          args.debug
-        );
-
-      const renderDurationSec = (s: number) =>
-        s >= 3600 ? `${Math.round(s / 3600)}h` : `${Math.round(s / 60)}m`;
-      // TODO: remove logging when we remove the launchdarkly file-transfer flag
-      if (args.debug) {
-        print2(`GET    (${renderDurationSec(expirySeconds.get)}): ${getUrl}`);
-        print2(
-          `DELETE (${renderDurationSec(expirySeconds.delete)}): ${deleteUrl}`
-        );
-      }
+      const s3 = createTransferClient(authn, target, args.debug);
 
       print2(`Uploading ${args.source}...`);
 
@@ -152,6 +138,25 @@ const fileTransferAction = async (
       }
 
       print2("Uploaded.");
+
+      // Sign the download/cleanup URLs only now that the file is uploaded — the
+      // GET window is finite, so we don't want it ticking during the upload.
+      const { getUrl, deleteUrl, expirySeconds } = await generateTransferUrls(
+        authn,
+        s3,
+        { bucket: target.bucket, key: uploadKey, awsSpec: target.awsSpec },
+        args.debug
+      );
+
+      const renderDurationSec = (s: number) =>
+        s >= 3600 ? `${Math.round(s / 3600)}h` : `${Math.round(s / 60)}m`;
+      // TODO: remove logging when we remove the launchdarkly file-transfer flag
+      if (args.debug) {
+        print2(`GET    (${renderDurationSec(expirySeconds.get)}): ${getUrl}`);
+        print2(
+          `DELETE (${renderDurationSec(expirySeconds.delete)}): ${deleteUrl}`
+        );
+      }
     },
     {
       command: "file-transfer",
--- a/src/commands/file-transfer.ts
+++ b/src/commands/file-transfer.ts
@@ -29,6 +29,9 @@ export type FileTransferCommandArgs = {
   debug?: boolean;
 };
 
+const renderDurationSec = (s: number) =>
+  s >= 3600 ? `${Math.round(s / 3600)}h` : `${Math.round(s / 60)}m`;
+
 export const fileTransferCommand = (yargs: yargs.Argv) =>
   yargs.command<FileTransferCommandArgs>(
     "file-transfer <source> <destination>",
@@ -148,8 +151,6 @@ const fileTransferAction = async (
         args.debug
       );
 
-      const renderDurationSec = (s: number) =>
-        s >= 3600 ? `${Math.round(s / 3600)}h` : `${Math.round(s / 60)}m`;
       // TODO: remove logging when we remove the launchdarkly file-transfer flag
       if (args.debug) {
         print2(`GET    (${renderDurationSec(expirySeconds.get)}): ${getUrl}`);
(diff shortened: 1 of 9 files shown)


src/plugins/file-transfer/__tests__/index.test.ts
Jenn738, line 1: i had the ai generate this test

src/plugins/file-transfer/__tests__/index.test.ts
```diff
@@ -0,0 +1,86 @@
...
+}));
+
+const ONE_HOUR = 60 * 60;
+const NOW = Date.parse("2030-01-01T00:00:00Z");
+
+describe("generateTransferUrls()", () => {
+  const target = {
+    bucket: "my-bucket",
+    key: "uploads/user/abc/file.txt",
+    awsSpec: {} as never,
```
MichaelDimitras: We use `as any` in tests more than `as never`.

I hardly see `as never` in the codebase.

src/plugins/file-transfer/index.ts
```diff
@@ -68,51 +68,78 @@ export const provisionTransferRequest = async (
   };
 };
 
+/**
+ * Builds an S3 client whose credentials refresh automatically. A large upload
+ * can run longer than the temporary credentials live; passing a provider
+ * function (that returns `expiration`) lets the SDK re-fetch fresh credentials
+ * mid-upload instead of failing the in-flight parts with ExpiredToken.
+ */
+export const createTransferClient = (
```
MichaelDimitras, line 78: If this is only used in `file-transfer.ts`, I think it makes sense to move it there.

src/commands/file-transfer.ts
```diff
@@ -152,6 +138,25 @@ const fileTransferAction = async (
...
+      // Sign the download/cleanup URLs only now that the file is uploaded — the
+      // GET window is finite, so we don't want it ticking during the upload.
+      const { getUrl, deleteUrl, expirySeconds } = await generateTransferUrls(
+        authn,
+        s3,
+        { bucket: target.bucket, key: uploadKey, awsSpec: target.awsSpec },
+        args.debug
+      );
+
+      const renderDurationSec = (s: number) =>
```
GGonryun: why isn't this a standalone function?

src/plugins/file-transfer/index.ts
```diff
@@ -68,51 +68,78 @@ export const provisionTransferRequest = async (
...
-  });
+  const { expiresAt } = await awsCloudAuth(authn, target.awsSpec, debug);
+  // If we somehow don't know the expiry, fall back to the configured limits
+  // (Infinity makes Math.min pick the constant). Math.max(0, ...) avoids signing
+  // an already-expired URL if the credentials are right at the edge.
+  const remaining =
+    expiresAt !== undefined
+      ? Math.max(0, Math.floor((expiresAt - Date.now()) / 1000))
+      : Infinity;
+  const getSeconds = Math.min(GET_EXPIRES_SECONDS, remaining);
```
GGonryun: tiny nit: variables that start with an action-verb make me think this is a function 🤔 ,

GGonryun (approved):

src/plugins/file-transfer/index.ts
```diff
@@ -68,51 +68,78 @@ export const provisionTransferRequest = async (
...
+  const deleteSeconds = Math.min(DELETE_EXPIRES_SECONDS, remaining);
 
   const objectArgs = { Bucket: target.bucket, Key: target.key };
   const [getUrl, deleteUrl] = await Promise.all([
     getSignedUrl(s3, new GetObjectCommand(objectArgs), {
-      expiresIn: GET_EXPIRES_SECONDS,
+      expiresIn: getSeconds,
```
MichaelDimitras: What happens if this is 0?  Will the user have to reauthenticate?

src/plugins/file-transfer/index.ts
```diff
@@ -68,51 +68,78 @@ export const provisionTransferRequest = async (
...
     }),
     getSignedUrl(s3, new DeleteObjectCommand(objectArgs), {
-      expiresIn: DELETE_EXPIRES_SECONDS,
+      expiresIn: deleteSeconds,
```
MichaelDimitras: I'm wondering about setting the expiry on the client side in general.  Theoretically someone could modify it.  I know there will be other controls on the backend, but its something I notice.

src/plugins/file-transfer/index.ts
```diff
@@ -68,51 +68,78 @@ export const provisionTransferRequest = async (
...
+  new S3Client({
+    region: target.region,
+    credentials: async () => {
+      const credentials = await awsCloudAuth(authn, target.awsSpec, debug);
+      return {
+        accessKeyId: credentials.AWS_ACCESS_KEY_ID,
+        secretAccessKey: credentials.AWS_SECRET_ACCESS_KEY,
+        sessionToken: credentials.AWS_SESSION_TOKEN,
+        // Providing `expiration` is what tells the SDK to call this provider
+        // again when the credentials are close to expiring. awsCloudAuth is
```
melissarh57:
```suggestion
        // again when the credentials are close to expiring. Aws SDK caches and refetches 
       
```

instead of our awsCloudAuth function maybe misleading

src/plugins/aws/assumeRole.ts
```diff
@@ -41,6 +41,7 @@ const stsAssume = async (
     AWS_SECRET_ACCESS_KEY: stsCredentials.SecretAccessKey,
     AWS_SESSION_TOKEN: stsCredentials.SessionToken,
     AWS_SECURITY_TOKEN: stsCredentials.SessionToken,
+    expiresAt: Date.parse(stsCredentials.Expiration), // epoch ms
```
melissarh57: do we want some kinda guard if aws gives us a non date? otherwise we'll have NaN not equal undefined but things could get weird in getSignedUrl({ expiresIn: NaN })). maybe if NaN fall back to undefined?

src/plugins/file-transfer/index.ts
melissarh57: that's a fair callout. I guess would we argue it is "protected" by the p0 file transfer request? and once they lose access to that the creds are invalid?
tho actually i wonder if the urls are still valid once the p0 file transfer permission is revoked

melissarh57: i think they aren't but that's accepted risk of using presigned urls. 

But ya to ur point Mike i think p0 granting the file transfer access makes the sts assuming the role and the expiration server side enforced? b/c if p0 then revokes they can't assume role anymore thru cli?

src/plugins/file-transfer/index.ts
Jenn738: are you suggesting this for all three functions? All three are only used in file-transfer.ts.

Jenn738: spoke offline, all functions in this file are helpers and only used in `file-transfer.ts` so decided to leave as is.

src/plugins/file-transfer/index.ts
Jenn738: spoke offline, but the STS credential lifetime should cap the URL. The cap should be whatever max_duration_session allows. This is defined within the IAM role itself.

src/plugins/file-transfer/index.ts
Jenn738: will add a warning

PR #419, opened 2026-06-19
Surface a clear OS Login hint on GCP SSH auth failures

MichaelDimitras:
## Problem

A `p0 ssh <instance> --provider gcloud` grant can succeed while the connection still fails when **OS Login is not enabled** in the customer's project — and P0 cannot enable it on their behalf. Without OS Login the user's key is never provisioned onto the VM (P0's grant doesn't include permission to write keys to instance metadata), so SSH auth is rejected with `Permission denied (publickey)`.

Today the user sees only that raw, generic rejection and concludes P0 is broken, when the real cause is a missing GCP prerequisite. (Customer-reported, P1.)

## Fix

Detect this specific failure and replace the generic error with an actionable message:

> Connected to `<instance>` but authentication was rejected (Permission denied (publickey)). The most common cause is OS Login not being enabled. Enable it by setting `enable-oslogin=TRUE` on the project (or instance) metadata, then retry. If OS Login is already enabled, this can also be a brief key-propagation delay or a just-granted IAM role — wait ~30s and retry. See <SSH-for-GCP prerequisites doc>.

It names OS Login as the **most likely** cause and lists the alternatives — it never claims OS Login is definitively off, since `Permission denied (publickey)` is generic.

## How it works

The CLI runs the system `ssh` with `gcloud compute start-iap-tunnel` as the `ProxyCommand`; both layers write to one stderr stream that `spawnSshNode` already captures.

- New pure classifier [`src/plugins/google/connection-error.ts`](https://github.com/p0-security/p0cli/blob/michaeldimitras/oslogin-validation/src/plugins/google/connection-error.ts): if the captured stderr contains `Permission denied (publickey)`, return the OS Login hint; otherwise return `undefined` (pass the raw error through).
- Wired in via a new optional `SshProvider.connectionErrorMessage` hook that `spawnSshNode` consults at terminal connection failure.
- **No new permissions.** The raw gcloud/ssh output stays available under `--debug`, and the exit code is preserved.

### Deliberately out of scope
We do **not** try to classify the other prerequisite failure (IAP / firewall not configured, which fails earlier at the gcloud tunnel). Its error strings vary by gcloud version and are easy to misattribute — and misattributing is worse than the status quo — so those failures pass through unchanged. A complementary proactive metadata check is being handled backend-side.

The existing ~2-minute propagation-retry window is unchanged: a publickey rejection is retried (covering genuine propagation delays) and the hint is surfaced when it ultimately times out.

## Testing
- Classifier unit tests: publickey → OS Login hint; OS Login wins when auth was reached even with earlier tunnel noise; tunnel/IAP error, unknown error, empty, and benign host-key warning → passthrough.
- Integration tests driving the real reject path via `sshProxy`: a publickey rejection surfaces the OS Login message end-to-end; an IAP/tunnel failure passes through with its exit code intact.
- Full suite green (195 tests, stable across repeated runs); `tsc`, eslint, prettier clean.
- The OS Login message was verified against a real instance with OS Login disabled.

## Note
This commit also bundles a small pre-existing WIP fix to the gcloud install package-selection command (+ its test) and a shared test `authn` helper, which were already in the working tree.

Screenshot 2026-06-18 at 5 34 55 PM

### DIFF
--- a/src/plugins/google/connection-error.ts
+++ b/src/plugins/google/connection-error.ts
@@ -1,0 +1,60 @@
+/** Copyright © 2024-present P0 Security
+
+This file is part of @p0security/cli
+
+@p0security/cli is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 3 of the License.
+
+@p0security/cli is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
+**/
+import { GcpSshRequest } from "./types";
+
+/**
+ * P0 grants the IAM roles needed for GCP SSH, but OS Login must be enabled in
+ * the customer's project — P0 cannot enable it on their behalf. When OS Login is
+ * off the IAM grant still succeeds, but the connection fails at SSH
+ * authentication: without OS Login the user's key is never provisioned onto the
+ * VM (P0's grant does not include permission to write keys to instance
+ * metadata), so auth is rejected with `Permission denied (publickey)`.
+ *
+ * Historically the user saw only that raw, generic rejection and concluded P0
+ * was broken. We surface a targeted hint instead. `Permission denied
+ * (publickey)` is not exclusively an OS Login problem — it can also be a brief
+ * key-propagation delay or a just-granted IAM role — so the message names OS
+ * Login as the most likely cause while listing the alternatives, and never
+ * claims certainty.
+ *
+ * We deliberately do NOT try to classify the other GCP prerequisite failure (IAP
+ * / firewall not configured, which fails earlier, at the gcloud tunnel rather
+ * than at SSH auth). Its `gcloud start-iap-tunnel` error strings vary by gcloud
+ * version and are easy to misattribute; since misattributing is worse than the
+ * status quo, those failures fall through to the raw error unchanged.
+ */
+
+export const GCP_SSH_PREREQUISITES_DOC =
+  "https://docs.p0.dev/integrations/resource-integrations/ssh#gcp-project-requirements";
+
+/** SSH auth was reached and rejected — most likely because OS Login is off. */
+const AUTH_REJECTED_PATTERN = /Permission denied \(publickey\)/;
+
+// Leads with a newline so it prints with one blank line above the preceding SSH
+// output, for legibility.
+const osLoginMessage = (instance: string) =>
+  `\nConnected to ${instance} but authentication was rejected ` +
+  `(Permission denied (publickey)). The most common cause is OS Login not ` +
+  `being enabled. Enable it by setting enable-oslogin=TRUE on the project (or ` +
+  `instance) metadata, then retry. If OS Login is already enabled, this can ` +
+  `also be a brief key-propagation delay or a just-granted IAM role — wait ` +
+  `~30s and retry. See ${GCP_SSH_PREREQUISITES_DOC}`;
+
+/**
+ * Inspects the captured stderr of a failed GCP SSH connection and returns an
+ * actionable message when the failure is an SSH auth rejection (most likely OS
+ * Login not being enabled), or `undefined` to fall through to the raw error.
+ */
+export const classifyGcpConnectionError = (
+  stderr: string,
+  request: Pick<GcpSshRequest, "id">
+): string | undefined =>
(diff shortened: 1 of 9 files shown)


src/plugins/google/install.ts
```diff
@@ -15,14 +15,14 @@ export const SupportedPlatforms = ["darwin"] as const;
...
-const GcpSshInstall: Readonly<Record<GcpSshItem, InstallMetadata>> = {
+export const GcpSshInstall: Readonly<Record<GcpSshItem, InstallMetadata>> = {
   gcloud: {
     label: "GCloud CLI",
     commands: {
       darwin: [
         // See https://cloud.google.com/sdk/docs/install-sdk
         "architecture=$(arch)",
-        'package=$([ $architecture = "arm64" ] && echo "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-cli-darwin-arm.tar.gz" || "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-cli-darwin-x86_64.tar.gz" )',
+        'package=$([ "$architecture" = "arm64" ] && echo "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-cli-darwin-arm.tar.gz" || echo "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-cli-darwin-x86_64.tar.gz" )',
```
GGonryun, line 25: i wonder if this independently fixes the install bug:

```
sudo p0 ssh sh-xxxxx--sudo --reason "test sudo"
 Password:
 Existing access found.  Connecting to instance.
 The following items must be installed on your system to continue:
   - GCloud CLI (gcloud)
 
 ? Do you want to install these automatically? (sudo access required) Yes
 
 Executing:
 architecture=$(arch) && \
 package=$([ $architecture = "arm64" ] && echo "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-cli-darwin-arm.tar.gz" || "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-cli-darwin-x86_64.tar.gz" ) && \
 wget -O ~/google-cloud-cli.tar.gz $package && \
 tar -xzf ~/google-cloud-cli.tar.gz -C ~ && \
 ~/google-cloud-sdk/install.sh && \
 rm -rf ~/google-cloud-cli.tar.gz && \
 sudo mkdir -p /usr/local/bin && \
 sudo ln -s ~/google-cloud-sdk/bin/gcloud /usr/local/bin/gcloud && \
 sudo chown root: /usr/local/bin/gcloud
 
 bash: line 1: https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-cli-darwin-x86_64.tar.gz: No such file or directory
 Shell exited with code 127
```

src/plugins/ssh/index.ts
```diff
@@ -272,7 +293,8 @@ async function spawnSshNode(
       if (!isAccessPropagated()) {
         if (options.endTime < Date.now()) {
           reject(
-            `Access did not propagate through ${provider.friendlyName} in time. ${getContactMessage()}`
+            connectionErrorMessage() ??
```
GGonryun: nit: i would prefer to have a knownError check, something like `const knownError = connectionErrorMessage()` and then `if (knownError) { reject(knownError) }` or return the generic error.

not a strong opinion tho.

src/plugins/google/ssh.ts
```diff
@@ -63,6 +64,9 @@ export const gcpSshProvider: SshProvider<
     return undefined;
   },
 
+  connectionErrorMessage: (stderr, request) =>
+    classifyGcpConnectionError(stderr, request),
```
GGonryun, line 68: nit, we could just inline this?

src/plugins/google/connection-error.ts
```diff
@@ -0,0 +1,60 @@
...
+const AUTH_REJECTED_PATTERN = /Permission denied \(publickey\)/;
+
+// Leads with a newline so it prints with one blank line above the preceding SSH
+// output, for legibility.
+const osLoginMessage = (instance: string) =>
+  `\nConnected to ${instance} but authentication was rejected ` +
+  `(Permission denied (publickey)). The most common cause is OS Login not ` +
+  `being enabled. Enable it by setting enable-oslogin=TRUE on the project (or ` +
+  `instance) metadata, then retry. If OS Login is already enabled, this can ` +
+  `also be a brief key-propagation delay or a just-granted IAM role — wait ` +
```
GGonryun, line 48: we should be retrying automatically to prevent key propagation delays.

src/plugins/google/connection-error.ts
GGonryun, line 1: my broader concern with this PR is it might be better to remove vm's that don't have oslogin from appearing on the list of ssh-able nodes instead. 🤔

GGonryun (approved):

PR #420, opened 2026-06-19
Melissahoo/cus 54 file transfer cli logic to invoke the download command

melissarh57:
**Problem**

The `file-transfer` command previously only handled half of the workflow: it uploaded a local file to a scoped S3 bucket and printed signed GET/DELETE URLs for the user to use manually. This change extends file-transfer to actually land the file on a target instance using a GET presigned url and `curl`.

**Change**

Extends `file-transfer` to complete the round trip in a single command:

1. Upload the local file to S3 (unchanged).
2. Request SSH access to the destination instance via the existing SSH approval flow.
3. Once approved, sign a fresh GET URL and have the instance pull the file via `curl` to the user's home directory.

The GET URL is now signed **after** the SSH approval clears rather than up front, so its 5-minute TTL covers the download window instead of being burned during the approval wait.

Also updates `delegation.aws` access to go through the typed `getDelegate` helper, matching how other plugins read delegation specs.

- [x] authentication / authorization (uses SSH approval flow for the download leg)
- [x] workflow (file-transfer now composes upload + SSH download)
- [ ] lifecycle SDK
- [ ] datastore abstractions

**Type of Change**

- [ ] Documentation update
- [ ] Dependency update
- [ ] Performance improvement
- [ ] Security fix

**Validation**

Manual end-to-end test against a dev environment:
- Uploaded a file with `file-transfer <source> <destination>`, confirmed S3 upload succeeded.
- Confirmed SSH approval prompt appeared for the destination instance.
- After approval, confirmed file landed at `/home/<linuxUser>/<basename>` on the target.
- Confirmed the command exits cleanly (no hangs from orphaned `session-manager-plugin` processes).

Error paths covered:
- Missing `curl` on the remote (exit code 127) surfaces a clear error pointing at the followup ticket for a fallback downloader.
- Non-zero remote exit codes propagate with the code in the error message.

Risks:
- The download command is hardcoded to `curl` for now. Most mainstream EC2 AMIs ship with curl, but air-gapped or minimal images may not. Tracked as followup.
- The forced `exitProcess(0)` mirrors what the SSH commands already do for the same orphaned-child-process reason.

**Tracking (optional)**
https://linear.app/p0-security/issue/CUS-54/file-transfer-cli-logic-to-invoke-the-download-command

**Implementation (optional)**

- Splits signed-URL generation: `generateSignedUrl` called to only sign the DELETE URL up front; `signGetUrl` is called later, after SSH approval, so the GET TTL isn't consumed by approval wait.
- The SSH leg reuses `prepareRequest` + `sshOrScp` from the existing SSH plumbing. `source` is stripped from args before reuse because `createCommand` routes on `"source" in args` to pick scp vs ssh, and we want the ssh branch here.
- `FileTransferPermissionSpec` moves `delegation.aws` from an intersection type into the `PermissionSpec` generic's delegation slot, matching the shape other permission specs use.

### DIFF
--- a/src/commands/file-transfer.ts
+++ b/src/commands/file-transfer.ts
@@ -89,8 +89,7 @@ const fileTransferAction = async (
       const target = await provisionTransferRequest(authn, args);
       print2(`Access approved for s3://${target.bucket}/${target.prefix}`);
 
-      // target.prefix is the backend-granted prefix (ends in `/`); append the
-      // local file's basename so the S3 object preserves the original filename.
+      // append original basename so the S3 object preserves the original filename.
       const uploadKey = `${target.prefix}${basename(args.source)}`;
 
       print2("Preparing upload credentials...");
@@ -159,14 +158,11 @@ const fileTransferAction = async (
 
       print2("Uploaded.");
 
-      // Step 2: have the destination instance pull the file from S3.
-      // This is a separate approval ticket (SSH provider) — file-transfer
-      // access granted bucket write only, not instance shell access.
+      // TODO we need to remove this second request. it should be included in file transfer delegation. Will be removed in future ticket
       print2(`Requesting download access on ${args.destination}...`);
 
       // Drop `source` (local file path) before passing to SSH plumbing —
-      // `createCommand` uses `"source" in args` to route between scp and ssh,
-      // and we want the ssh branch here.
+      // `createCommand` uses `"source" in args` to branch between scp and ssh path, and we want the ssh branch here.
       const { source: _source, ...sshBaseArgs } = args;
       const sshCmdArgs = {
         ...sshBaseArgs,
@@ -177,7 +173,7 @@ const fileTransferAction = async (
       const { request, requestId, privateKey, sshProvider, sshHostKeys } =
         await prepareRequest(authn, sshCmdArgs, args.destination);
 
-      // Re-sign GET URL now so the 5-min TTL starts after approval clears,
+      // Sign GET URL now so the 5-min TTL starts after approval clears,
       // not before — otherwise long approval waits could expire the URL.
       const { signedUrl: getUrl, expirySeconds: getExpirySeconds } =
         await generateSignedUrl(
--- a/src/commands/file-transfer.ts
+++ b/src/commands/file-transfer.ts
@@ -14,9 +14,8 @@ import { print2 } from "../drivers/stdio";
 import { exitProcess, traceSpan } from "../opentelemetry/otel-helpers";
 import {
   createTransferClient,
-  generateTransferUrls,
+  generateSignedUrl,
   provisionTransferRequest,
-  signGetUrl,
 } from "../plugins/file-transfer";
 import { sshOrScp } from "../plugins/ssh";
 import { prepareRequest } from "./shared/ssh";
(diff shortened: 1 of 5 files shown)


src/commands/file-transfer.ts
```diff
@@ -142,21 +158,73 @@ const fileTransferAction = async (
...
+        request,
+        requestId,
+        cmdArgs: downloadCmdArgs,
+        privateKey,
+        sshProvider,
+        sshHostKeys,
+      });
+
+      if (exitCode === 127) {
+        throw `curl not found on ${args.destination}. The file is in S3 — install curl on the instance and re-run, or wait for CUS-68 to add a fallback downloader.`;
```
GGonryun: nit on this error message, if this is public we should throw something that is user actionable.

src/commands/file-transfer.ts
```diff
@@ -142,21 +158,73 @@ const fileTransferAction = async (
...
+        `Downloading to ${request.linuxUserName}@${args.destination}:${remotePath}...`
       );
 
-      // TODO: remove logging when we remove the launchdarkly file-transfer flag
-      if (args.debug) {
-        print2(`GET    (${renderDurationSec(expirySeconds.get)}): ${getUrl}`);
-        print2(
-          `DELETE (${renderDurationSec(expirySeconds.delete)}): ${deleteUrl}`
-        );
+      // TODO(CUS-68): final download command (curl vs aws s3 cp vs wget) is
```
GGonryun: nit, it's better not to reference tickets in code in case the ticket moves or changes or something, we can just keep the TODO comment.

src/commands/file-transfer.ts
```diff
@@ -142,21 +158,73 @@ const fileTransferAction = async (
...
-        authn,
-        s3,
-        { bucket: target.bucket, key: uploadKey, awsSpec: target.awsSpec },
-        args.debug
+      // TODO we need to remove this second request. it should be included in file transfer delegation. Will be removed in future ticket
+      print2(`Requesting download access on ${args.destination}...`);
+
+      // Drop `source` (local file path) before passing to SSH plumbing —
+      // `createCommand` uses `"source" in args` to branch between scp and ssh path, and we want the ssh branch here.
+      const { source: _source, ...sshBaseArgs } = args;
```
GGonryun, line 166: nit, if we have lodash can we use `omit`?

src/plugins/file-transfer/index.ts
```diff
@@ -127,29 +128,23 @@ export const generateTransferUrls = async (
         `Check your system clock or re-run the request.`
     );
   }
-  const secondsToExpireGetUrl = Math.min(SECONDS_TO_EXPIRE_GET_URL, remaining);
-  const secondsToExpireDeleteUrl = Math.min(
-    SECONDS_TO_EXPIRE_DELETE_URL,
-    remaining
-  );
+  const maxExpiry =
+    command === "get"
```
GGonryun: i might be overthinking this but i generally like using maps for these kinds of checks because they are more exhaustive:

```
type SignedUrlCommand = "delete" | "get"
const URL_GENERATORS: Record<SignedUrlCommand, ...> = { 
  "get": { expiresIn: MAX_SECONDS_TO_EXPIRE_GET_URL, command: new GetObjectCommand({ Bucket: target.bucket, Key: target.key })  
  "delete": { expiresIn: MAX_SECONDS_TO_EXPIRE_DELETE_URL, command: new DeleteObjectCommand({ Bucket: target.bucket, Key: target.key })
}
```

im happy with what you have now too, just a tiny stylistic nit.

GGonryun (approved):

src/commands/file-transfer.ts
melissarh57: tried to do with omit but b/c the type yargs.ArgumentsCamelCase has [argName: string]: unknown; the type system gets weird with omit, vs works fine with destructuring

melissarh57: image

src/commands/file-transfer.ts
melissarh57: will update!

src/commands/file-transfer.ts
melissarh57: will update!

src/plugins/file-transfer/index.ts
melissarh57: will update!

PR #423, opened 2026-06-23
File Transfer: Delete object from s3 bucket after download complete

Jenn738:
**Problem**

File Transfer integration did not delete the object after download has successfully completed. This change uses the delete presigned url to delete the object from the S3 bucket after the object has been successfully downloaded into the EC2 instance.

- [ ] workflow
- [ ] lifecycle SDK
- [ ] datastore abstractions

**Type of Change**

- [ x] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] Dependency update
- [ ] Performance improvement
- [ ] Security fix

**Validation**

Verified that object was no longer in the S3 bucket after download completed.

### DIFF
--- a/src/commands/file-transfer.ts
+++ b/src/commands/file-transfer.ts
@@ -12,16 +12,14 @@ import { retryWithSleep } from "../common/retry";
 import { authenticate } from "../drivers/auth";
 import { print2 } from "../drivers/stdio";
 import { exitProcess, traceSpan } from "../opentelemetry/otel-helpers";
-import { AwsResourcePermissionSpec } from "../plugins/aws/types";
 import {
   createTransferClient,
   generateSignedUrl,
   provisionTransferRequest,
 } from "../plugins/file-transfer";
 import { sshOrScp } from "../plugins/ssh";
-import { Authn } from "../types/identity";
 import { prepareRequest } from "./shared/ssh";
-import { S3Client } from "@aws-sdk/client-s3";
+import { DeleteObjectCommand, S3Client } from "@aws-sdk/client-s3";
 import { Upload } from "@aws-sdk/lib-storage";
 import { createReadStream, statSync } from "fs";
 import { basename } from "node:path";
@@ -42,11 +40,9 @@ const COMMAND_NOT_FOUND_EXIT_CODE = 127;
 const SUCCESS_EXIT_CODE = 0;
 
 /**
- * Best-effort cleanup of the uploaded S3 object via a presigned DELETE URL.
- * Call this only after a confirmed-successful download.
- *
- * The DELETE URL is signed here, at the point of use, rather than up front — a
- * long approval wait would otherwise burn its TTL before we ever issue it.
+ * Best-effort cleanup of the uploaded S3 object. The CLI already holds an
+ * authenticated S3 client, so it deletes directly and the client also
+ * auto-refreshes credentials.
  *
  * This must never fail an otherwise-successful transfer: the object still
  * expires via the bucket's lifecycle policy, so a failed delete is harmless.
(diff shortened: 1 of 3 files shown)


src/commands/file-transfer.ts
```diff
@@ -34,6 +37,53 @@ export type FileTransferCommandArgs = {
...
+ */
+const deleteUploadedObject = async (
+  authn: Authn,
+  s3: S3Client,
+  target: { bucket: string; awsSpec: AwsResourcePermissionSpec },
+  key: string,
+  debug?: boolean
+) => {
+  try {
+    const { signedUrl, expirySeconds } = await generateSignedUrl(
```
GGonryun: question, maybe as a separate ticket but we do have an audit log endpoint that ssh uses to submit information about what the CLI is doing to the p0 backend, can we instrument this w/ audit logs so we can detect when clean up fails?

src/commands/file-transfer.ts
```diff
@@ -34,6 +37,53 @@ export type FileTransferCommandArgs = {
...
+      debug
+    );
+    if (debug) {
+      print2(`DELETE (${renderDurationSec(expirySeconds)}): ${signedUrl}`);
+    }
+    const response = await fetch(signedUrl, { method: "DELETE" });
+    if (!response.ok) {
+      throw new Error(`S3 returned ${response.status} ${response.statusText}`);
+    }
+    print2(`Deleted s3://${target.bucket}/${key} from the bucket.`);
```
GGonryun: this feels like a debug message, we can just end with "File transfer succeeded" and not report on the intricacies of the cleanup

GGonryun (approved):

src/commands/file-transfer.ts
```diff
@@ -34,6 +37,53 @@ export type FileTransferCommandArgs = {
...
+  try {
+    const { signedUrl, expirySeconds } = await generateSignedUrl(
+      authn,
+      s3,
+      { bucket: target.bucket, key, awsSpec: target.awsSpec },
+      "delete",
+      debug
+    );
+    if (debug) {
+      print2(`DELETE (${renderDurationSec(expirySeconds)}): ${signedUrl}`);
```
melissarh57: i think we want to keep the todo comment to remove before going live to avoid cred logging?

src/commands/file-transfer.ts
```diff
@@ -34,6 +37,53 @@ export type FileTransferCommandArgs = {
...
+    }
+    const response = await fetch(signedUrl, { method: "DELETE" });
+    if (!response.ok) {
+      throw new Error(`S3 returned ${response.status} ${response.statusText}`);
+    }
+    print2(`Deleted s3://${target.bucket}/${key} from the bucket.`);
+  } catch (err) {
+    if (debug) {
+      const message = err instanceof Error ? err.message : String(err);
+      print2(`Warning: failed to delete S3 object: ${message}`);
```
melissarh57: can we add another print to log which object failed to delete? and say: you may delete this object manually or it will get picked up by the file transfer s3 24 lifecycle expiration rule?

melissarh57: also, i'm thinking the technical error shoudl go under debug but the error that failed to delete s3 object but explain it's fine and which object maybe shoudn't be under debug b/c it's useful info they may want to know about the outcome

melissarh57 (commented):
claude and i realized now: we don't actually need a presigned delete url. we could just use the s3 client to delete the object directly instead. since it's already authed to the bucket. it may be worth the switch now in this PR, wouldn't be that hard? just need to have the created s3 client available in this main method?

don't worry about ripping out the "delete" in generateSignedUrl we could do that later. what do you think?

src/commands/file-transfer.ts
Jenn738: There is an audit logs ticket, but I haven't worked on it yet. That's up next! I'll make a note of this though.

src/commands/file-transfer.ts
Jenn738: changed the approach, so this seems no longer relevant

src/commands/file-transfer.ts
Jenn738: changed to debug!

src/commands/file-transfer.ts
```diff
@@ -34,6 +35,46 @@ export type FileTransferCommandArgs = {
...
+    await s3.send(new DeleteObjectCommand({ Bucket: bucket, Key: key }));
+    if (debug) {
+      print2(`Deleted s3://${bucket}/${key} from the bucket.`);
+    }
+  } catch (err) {
+    print2(
+      `Warning: could not delete s3://${bucket}/${key}. The file transfer succeeded, 
+      so this is safe to ignore. You may delete this object manually, or it will be 
+      removed automatically by the file-transfer bucket's lifecycle expiration 
+      rule.`
```
melissarh57: nit: may want to clean up new lines and whitespace

src/commands/file-transfer.ts
```diff
@@ -34,6 +35,46 @@ export type FileTransferCommandArgs = {
...
+ * Best-effort cleanup of the uploaded S3 object. The CLI already holds an
+ * authenticated S3 client, so it deletes directly and the client also
+ * auto-refreshes credentials.
+ *
+ * This must never fail an otherwise-successful transfer: the object still
+ * expires via the bucket's lifecycle policy, so a failed delete is harmless.
+ * The result therefore defaults to success — errors are only surfaced under
+ * --debug.
+ */
+const deleteUploadedObject = async (
```
melissarh57, line 52: can you add a todo in generateSignedUrl to remove delete path and remove delete tests?

melissarh57 (approved):
looks good just 2 comments! thanks!

src/commands/file-transfer.ts
Jenn738: oops

melissarh57 (approved):
awesome thanks!

PR #438, opened 2026-07-08
create a shared file for common azure ssh functions

GGonryun:
**Problem**

The Azure Bastion SSH provider (`ssh-bastion.ts`) previously contained a mix of Azure-specific plumbing — login handling, install checks, key generation, sudo-propagation detection, and repro-command construction — inline in the provider definition. As this codebase adds more ways to reach an Azure host (e.g. jump hosts), that logic needs to be reused by other Azure SSH providers rather than duplicated or copy-pasted.

**Change**

This PR extracts the pieces of `ssh-bastion.ts` that are not specific to Bastion tunneling into `src/plugins/azure/ssh-shared.ts`, so future Azure SSH providers can reuse them:

- `azureSshProviderBase`: an `SshProvider` partial with the shared `cloudProviderLogin`, `ensureInstall`, `friendlyName`, `loginRequiredMessage`, `propagationTimeoutMs`, `preTestAccessPropagationArgs`, `generateKeys`, access-provisioning patterns, and `toCliRequest`. `ssh-bastion.ts` now spreads this (`...azureSshProviderBase`) and only defines the Bastion-specific `proxyCommand`, `reproCommands`, and `setup`.
- `azureSshLoginReproCommands`: the shared `az login` / `az account set` / cert-generation repro command sequence, which `ssh-bastion.ts` now appends its tunnel command to.

To avoid a require cycle (`ssh-shared.ts` → `commands/shared/ssh.ts` → `ssh-bastion.ts`, which spreads `azureSshProviderBase` at module load time), `ssh-shared.ts` imports `isSudoCommand` as a value but `SshAdditionalSetup` as a type-only import.

No behavior change — this is a pure extraction/refactor.

- [ ] workflow
- [ ] lifecycle SDK
- [ ] datastore abstractions

**Type of Change**

- [ ] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] Dependency update
- [ ] Performance improvement
- [ ] Security fix

**Validation**

Expanded `src/plugins/azure/__tests__/ssh-shared.test.ts` with coverage for the newly-extracted `azureSshLoginReproCommands` (including the identity-file-present vs. home-directory-fallback key path cases). Ran the full suite for this file — 15/15 tests pass. `tsc --noEmit` is clean. No behavior changes were made, so existing `ssh-bastion` tests continue to cover the composed provider.

**Tracking (optional)**

CUS-647

**Implementation (optional)**

The main technical decision was the type-only import of `SshAdditionalSetup` in `ssh-shared.ts` to break a require cycle back through `commands/shared/ssh.ts`, since `ssh-bastion.ts` spreads `azureSshProviderBase` at module load time.

**Attachments (optional)**

N/A

**Follow-up Work (optional)**

`azureSshProviderBase` and `azureSshLoginReproCommands` are intended to be reused by an upcoming jump-host Azure SSH provider.

### DIFF
--- a/src/commands/shared/__tests__/ssh-resolve.test.ts
+++ b/src/commands/shared/__tests__/ssh-resolve.test.ts
@@ -10,6 +10,7 @@ You should have received a copy of the GNU General Public License along with @p0
 **/
 import { authenticate } from "../../../drivers/auth";
 import { awsSshProvider } from "../../../plugins/aws/ssh";
+import { TARGET_CONNECT_TIMEOUT_SECONDS } from "../../../plugins/azure/ssh-jump-host";
 import { Authn } from "../../../types/identity";
 import { sshResolveAction } from "../../ssh-resolve";
 import { SshResolveCommandArgs } from "../ssh";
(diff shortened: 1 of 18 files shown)


smanzler (approved):
lgtm

gusboling (approved):

PR #444, opened 2026-07-16
Surface an actionable hint when the Azure CLI ssh extension is missing

MichaelDimitras:
**Problem**

Azure SSH access through the P0 CLI runs `az ssh cert`, which requires the Azure CLI's `ssh` extension. The extension is not part of the base Azure CLI install; when it is missing, the Azure CLI tries to install it on first use, and that dynamic install can fail — most commonly at the pip step, in environments where the CLI cannot write to its own extension directory. A user hitting this (user-reported issue) sees only the raw Azure CLI output followed by a generic error, and reasonably concludes the P0 CLI is broken:

```
WARNING: The command requires the extension ssh. It will be installed first.
ERROR: An error occurred. Pip failed with status code 1. Use --debug for more information.

Failed to generate Azure AD SSH certificate: Error: Sub-process exited with code
```

**Change**

Two halves: stop the user reaching that state, and explain it when they do anyway.

*Check for the extension before running anything.* `ensureAzInstall` already confirms the Azure CLI is present. It now also confirms the `ssh` extension is, and stops with install instructions if it isn't, rather than letting `az ssh cert` fail later. The extension is not a standalone binary, so it cannot be found with `which` the way every other tool here is — it is probed by running `az extension show --name ssh` and seeing whether that returns anything.

A fresh install no longer lands without it either: on macOS the Azure CLI's install commands end with `az extension add --name ssh`, so the CLI and the extension arrive together.

*Explain the failure if it still happens.* The pre-flight check does not cover every route to a broken extension — an extension present but unusable, a machine where the CLI was installed outside our install path — so the failure path is still handled, following the same pattern as the GCP OS Login hint (#419):

> Failed to generate an Azure AD SSH certificate because the Azure CLI's 'ssh' extension is not installed, and it could not be installed automatically. Install it manually by running 'az extension add --name ssh', then retry this command. If that install fails, update the Azure CLI to the latest version and retry it.

- New pure classifier `src/plugins/azure/cert-error.ts`: inspects the captured output of a failed `az ssh cert` and returns the hint only when the `ssh` extension is implicated; otherwise returns `undefined` and the raw error passes through unchanged.
- Wired into the `catch` in `generateSshKeyAndAzureAdCert` (`ssh-shared.ts`). The raw Azure CLI output is still printed before the hint, and all other failure modes are unaffected.
- The remediation wording is one function in `cert-error.ts`, used by both the pre-flight message and the error hint, so the two cannot drift apart.

The classifier is deliberately conservative — misattributing an unrelated failure to the extension would be worse than the status quo. The Azure CLI's "The command requires the extension ssh" warning alone is benign (when the auto-install succeeds the command continues and exits zero), so it does not trigger the hint by itself. The hint triggers only when:

- that warning is accompanied by a failed install (`Pip failed with status code ...`), or
- the `ssh` command group doesn't resolve at all (`'ssh' is not in the 'az' command group` / `'ssh' is misspelled or not recognized by the system`), which happens when dynamic install is disabled or unsupported.

**Type of Change**

**Validation**

Automated tests are included:

- Classifier unit tests (`cert-error.test.ts`): non-interactive and interactive pip-failure transcripts → hint; unresolved `ssh` command group (both wordings) → hint; auto-install-succeeded-then-unrelated-failure, unrelated login failure, and empty output → passthrough.
- Wiring test (`ssh-shared.test.ts`): a rejected `az ssh cert` exec with the missing-extension stderr rejects with the hint end-to-end; the existing test pins that unrelated failures keep the original generic message.
- Pre-flight tests (`install.test.ts`): both tools present → succeeds; Azure CLI missing → fails without probing the extension; extension missing → prints the instructions and fails; and the macOS install command list ends with the extension add, so the CLI is never installed alone.

`tsc`, eslint and prettier are clean, and CI's `Test` job is green at `cce35258`. Tests were written first and observed failing before the implementation was added.

The failure path of `az ssh cert` is unchanged where the classifier does not fire — byte for byte. What is new on the happy path is the pre-flight probe: every Azure SSH command now runs one extra `az extension show` before doing any work. That is a subprocess round trip per invocation, and it is also a new way for a working setup to be refused — if `az extension show` fails for a reason unrelated to the extension being absent, the command stops with the install message instead of proceeding.

**Tracking (optional)**

https://linear.app/p0-security/issue/CUS-768/better-error-handling-when-the-az-ssh-extension-is-not-installed

**Follow-up Work (optional)**

- Other dynamic-install failure modes (e.g. network errors while downloading the extension) intentionally pass through unclassified; their wording varies by Azure CLI version.
- The P0 docs' Azure SSH prerequisites page could mention the client-side `ssh` extension requirement.
- Only the macOS entry has install commands, so the extension add rides along on macOS only. Elsewhere the pre-flight check reports what is missing and the user installs it themselves, which is what the Azure CLI itself already did there.
- The probe runs on every Azure SSH invocation. If that round trip shows up, its result could be cached for the life of the process.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/plugins/azure/ssh-shared.ts
```diff
@@ -83,7 +84,10 @@ export const generateSshKeyAndAzureAdCert = async (
   } catch (error: any) {
     print2(error.stdout);
     print2(error.stderr);
-    throw `Failed to generate Azure AD SSH certificate: ${error}`;
+    const hint = classifyAzureCertGenerationError(
+      `${error.stdout ?? ""}\n${error.stderr ?? ""}`
+    );
+    throw hint ?? `Failed to generate Azure AD SSH certificate: ${error}`;
```
MichaelDimitras: Looking at this, I'm wondering about just passing `Failed to generate Azure AD SSH certificate: ${error}` to `classifyAzureCertGenerationError` as a fallback. It seems not obvious that `classifiy` returning undefined should require a message to be created here.

src/plugins/azure/install.ts
```diff
@@ -9,27 +9,57 @@ This file is part of @p0security/cli
...
+  (await checkToolVersion(
+    "the Azure CLI 'ssh' extension",
+    ["az", "extension", "show", "--name", "ssh"],
+    debug
+  )) !== undefined;
+
+export const ensureAzInstall = async (debug?: boolean) => {
+  if (!(await ensureInstall(AzItems, AzInstall))) return false;
+
+  if (await isAzSshExtensionInstalled(debug)) return true;
```
MichaelDimitras: In practice, I think these `ensureInstall` functions are only used for ssh, but if a different Azure integration ever uses this, we'll wrongly require ssh to be installed when it doesn't apply.  Maybe in the case of RDP for instance.

src/plugins/azure/cert-error.ts
```diff
@@ -0,0 +1,60 @@
...
+];
+
+export const AZ_SSH_EXTENSION_ADD_COMMAND = "az extension add --name ssh";
+
+const azSshExtensionPipHint = (
+  os: ReturnType<typeof getOperatingSystem> = getOperatingSystem()
+) =>
+  os === "win"
+    ? `If that command fails with a pip error, update the Azure CLI:\n\n` +
+      `  az upgrade\n`
```
MichaelDimitras: You don't need to run the `python3 -m ensurepip --upgrade` command on Windows?

src/plugins/azure/install.ts
gusboling: Yeah I think this makes sense; especially since it might seem suspicious from a security perspective if we require SSH functionality for seemingly no reason

src/plugins/azure/ssh-shared.ts
gusboling: I also prefer this.

gusboling (approved):

src/plugins/azure/cert-error.ts
GGonryun: i forgot to remove this, turns out i don't need this in general

PR #449, opened 2026-07-21
add file-transfer e2e tests

GGonryun:
**Problem**

`p0 file-transfer` is a recently added command that moves a local file onto an AWS instance through a temporary S3 bucket, but nothing exercised it end to end — the existing e2e suite only covers the ssh flows. Regressions in the transfer path (access provisioning, S3 upload, remote download) would only surface when a user ran the command against real infrastructure.

**Change**

Adds a file-transfer e2e suite, scaffolded on the existing ssh e2e conventions. Since the command only supports AWS destinations today, it is a single spec (`src/e2e/file-transfer-aws.e2e.ts`) targeting the same node as the AWS ssh flow (`P0_E2E_AWS_NODE`), running in order:

1. `p0 ls file-transfer session destination --json` — confirms the node is visible before requesting access
2. `p0 file-transfer` with a nonexistent source — asserts it fails fast, before any access request
3. `p0 file-transfer` with a marker file — asserts the full upload/approval/download path succeeds
4. `p0 ssh ... cat <remote-path>` — verifies the content actually landed on the instance, reusing the grant from step 3

The ls-destination matching logic is extracted from `ssh-flow.ts` into the shared harness so both suites use it, and the e2e README documents the new flow.

- [ ] workflow
- [ ] lifecycle SDK
- [ ] datastore abstractions

**Type of Change**

- [ ] Documentation update
- [ ] Dependency update
- [ ] Performance improvement
- [ ] Security fix

**Validation**

`yarn lint:types`, `yarn lint:code`, and `prettier --check src/e2e/` all pass. The new spec is only picked up by the e2e config (`src/e2e/**/*.e2e.ts`), so the unit test run (`**/*.test.ts`) is unaffected; no production code is touched.

The suite itself has not yet been run against the live e2e org (it requires a login session, a configured AWS node, and auto-approved access) — that first run is the remaining validation step.

**Tracking (optional)**

CUS-602

**Implementation (optional)**

- Kept the suite as one self-contained spec instead of the flow/spec split the ssh suite uses, since only one provider is supported; a comment notes to split like `ssh-flow.ts` if more providers arrive.
- Step 1 assumes the backend supports `p0 ls file-transfer session destination`, mirroring how the provision request structures its arguments; if the backend doesn't implement that listing, that step is the one to adjust.

**Attachments (optional)**

N/A

### DIFF
--- a/src/e2e/README.md
+++ b/src/e2e/README.md
@@ -6,19 +6,22 @@ Automated end-to-end tests that exercise the real CLI against a live P0 organiza
 yarn e2e
 ```
 
-Target a single cloud provider's flow by passing its spec file (see "What it does" below for the file names):
+Target a single flow by passing its spec file (see "What it does" below for the file names):
 
 ```
 yarn e2e src/e2e/ssh-aws.e2e.ts
 yarn e2e src/e2e/ssh-azure.e2e.ts
 yarn e2e src/e2e/ssh-gcloud.e2e.ts
+yarn e2e src/e2e/file-transfer-aws.e2e.ts
 ```
 
-Global setup and login still run first; only that provider's node needs to be configured (see Configuration below).
+Global setup and login still run first; only that flow's node needs to be configured (see Configuration below).
 
 ## What it does
 
 Global setup (before any test) smoke-tests the existing build with `p0 --version` and runs `p0 login p0-e2e` — a failure of either step aborts the whole run. Run `yarn build` yourself first; the suite does not build for you.
+
+### The ssh flow
 
 There's one ssh flow spec per cloud provider — `ssh-aws.e2e.ts`, `ssh-azure.e2e.ts`, `ssh-gcloud.e2e.ts` — each driving its configured node through the whole access lifecycle, in order. They share their step logic (`ssh-flow.ts`) but are separate files so you can run one provider independently, e.g. `yarn e2e src/e2e/ssh-aws.e2e.ts`. Each step runs a remote command instead of an interactive shell, so its session ends on its own once the connection succeeds:
 
@@ -36,6 +39,17 @@ Include <P0_PATH>/ssh/configs/*.config
 ```
 
 (`<P0_PATH>` is `~/.p0` or `~/.p0-<env>` depending on your CLI environment; extra `ssh-resolve` flags such as `--debug` are fine, but the `Match` line must come before the `Include`.)
+
+### The file-transfer flow
+
+`file-transfer-aws.e2e.ts` is a separate suite for `p0 file-transfer`, which only supports AWS destinations today. It targets the same AWS node as the ssh flow (`P0_E2E_AWS_NODE`) and runs, in order:
+
+1. `p0 ls file-transfer session destination --json` — confirms the configured node is visible to the e2e user before requesting access
+2. `p0 file-transfer <missing-file> <node-id>` — asserts a nonexistent source path fails fast, before any access request is made
+3. `p0 file-transfer <file> <node-id>` — uploads a marker file through the temporary S3 bucket, waits for approval and propagation, and downloads it onto the instance (this implicitly provisions ssh access to the node)
+4. `p0 ssh <node-id> cat <remote-path>` — verifies the file content actually landed on the instance, reusing the grant from step 3
+
+The suite removes the remote file (best effort) when it finishes, logging a warning if that cleanup fails. Run it together with the ssh flows or on its own — each suite requests its own access.
 
 At the end of the run the suite prints a reminder: **please revoke all access granted to the e2e user before running the suite again**, so the next run requests access from scratch.
 
--- a/src/e2e/file-transfer-aws.e2e.ts
+++ b/src/e2e/file-transfer-aws.e2e.ts
@@ -1,0 +1,159 @@
+/** Copyright © 2024-present P0 Security
+
+This file is part of @p0security/cli
+
+@p0security/cli is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, version 3 of the License.
+
+@p0security/cli is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with @p0security/cli. If not, see <https://www.gnu.org/licenses/>.
+**/
+import {
+  AWS_TARGET,
+  E2E_REASON,
+  LsDestinationItem,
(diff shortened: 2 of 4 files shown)


src/e2e/file-transfer-aws.e2e.ts
```diff
@@ -0,0 +1,152 @@
...
+          provider,
+          "--reason",
+          E2E_REASON,
+          "rm",
+          "-f",
+          remotePath,
+        ],
+        { timeoutMs: 5 * 60_000 }
+      );
+    }
```
Jenn738, line 69: nit: not necessary, but is it valuable to log if remote cleanup hasn't happened since we don't and shouldn't have any asserts in this section?

src/e2e/file-transfer-aws.e2e.ts
```diff
@@ -0,0 +1,152 @@
...
+        "file-transfer",
+        "session",
+        "destination",
+        "--json",
+        "--size",
+        "500",
+      ],
+      { timeoutMs: 2 * 60_000 }
+    );
+    expect(result.code, result.output).toBe(0);
```
Jenn738: mostly a question/nit: I assume 0 is success? do we have constants defined anywhere for status codes that we could use here?

src/e2e/file-transfer-aws.e2e.ts
```diff
@@ -0,0 +1,152 @@
...
+        "--reason",
+        E2E_REASON,
+      ],
+      { timeoutMs: 2 * 60_000 }
+    );
+    expect(result.code, result.output).not.toBe(0);
+    expect(result.output).toContain("Source file not found");
+  });
+
+  it("requests access with p0 file-transfer and puts the file on the instance", async () => {
```
Jenn738, line 116: how long is the timeout on this by default? Would it make sense to define a narrower timeout bound given that this is an e2e test?

src/e2e/README.md
```diff
@@ -37,6 +40,17 @@ Include <P0_PATH>/ssh/configs/*.config
...
+### The file-transfer flow
+
+`file-transfer-aws.e2e.ts` is a separate suite for `p0 file-transfer`, which only supports AWS destinations today. It targets the same AWS node as the ssh flow (`P0_E2E_AWS_NODE`) and runs, in order:
+
+1. `p0 ls file-transfer session destination --json` — confirms the configured node is visible to the e2e user before requesting access
+2. `p0 file-transfer <missing-file> <node-id>` — asserts a nonexistent source path fails fast, before any access request is made
+3. `p0 file-transfer <file> <node-id>` — uploads a marker file through the temporary S3 bucket, waits for approval and propagation, and downloads it onto the instance (this implicitly provisions ssh access to the node)
+4. `p0 ssh <node-id> cat <remote-path>` — verifies the file content actually landed on the instance, reusing the grant from step 3
+
+The suite removes the remote file (best effort) when it finishes. Run it together with the ssh flows or on its own; when both AWS suites run in one invocation, the second one reuses the access granted by the first.
```
Jenn738: Is the reuse actually happening? Should there be a test that shows this happens?

Jenn738 (approved):

src/e2e/harness.ts
```diff
@@ -65,6 +65,31 @@ export const GCLOUD_TARGET = sshTarget(
...
 );
 
+/** Item shape of `p0 ls <command> session destination --json` output. */
+export type LsDestinationItem = {
+  provider: string;
+  key: string;
+  value: string;
+  instanceId?: string;
+  name?: string;
+  alternativeNames?: string[];
```
Jenn738, line 84: this is mostly a question that doesn't block this PR, but what is the name and alternativeNames fields used for?

Jenn738 (approved):

src/e2e/file-transfer-aws.e2e.ts
GGonryun: since we only run these locally i don't mind using a longer time out

src/e2e/harness.ts
GGonryun: the name/alternativeNames are how we are able to target instances using `p0 ssh <name>`, its a relatively new abstraction that will make it easier for customers too to customize what identifiers they want to use to target their nodes.

src/e2e/file-transfer-aws.e2e.ts
GGonryun: Good call — the cleanup now captures the result and logs an `[e2e] warning:` line (with the command output) when the remote `rm` exits non-zero. Still best-effort, no asserts. Fixed in bc5df11.

src/e2e/file-transfer-aws.e2e.ts
GGonryun: Yes — 0 is the standard POSIX success exit code. There wasn't a constant the e2e suite could reach (`src/commands/file-transfer.ts` has a private `SUCCESS_EXIT_CODE`), so I added one to the e2e harness and switched this spec and `ssh-flow.ts` to use it. Fixed in bc5df11.

src/e2e/file-transfer-aws.e2e.ts
GGonryun: By default `runP0` kills the process after 15 minutes, inside vitest's 20-minute per-test timeout (see `vitest.e2e.config.mts`). This is the step that requests access, and ssh access propagation alone can take up to ~10 minutes (plus the S3 IAM policy propagation retries during upload), so a narrower bound here would mostly buy us flakes. The steps that don't wait on propagation — the `p0 ls` lookup and the missing-source check — are already bounded to 2 minutes.

src/e2e/README.md
GGonryun: Fair challenge — the suite doesn't assert that, and whether the backend reuses an existing grant across the two request types (ssh session vs. file-transfer session) is backend behavior this suite shouldn't encode. I removed the claim; the README now just says each suite requests its own access (bc5df11). Even within this suite, the verify step's `p0 ssh` would pass via a fresh auto-approved request whether or not the transfer's delegated ssh grant is reused, so proving reuse would need backend-side assertions — I'd treat that as backend test territory.

GGonryun: its sort of happening `within` a test, if you look at like the AWS test suite, we only ask for access at the start of the test and then continue along doing `ssh/scp`

src/e2e/harness.ts
GGonryun: They're extra identifiers the backend includes on `p0 ls ... destination --json` items beyond the `key`/`value` listing pair — the instance's display name and any other names the node is known by (e.g. alternative hostnames). `p0 ssh` accepts any of these as the destination, so the matcher accepts all of them — that way `P0_E2E_*_NODE` can be set to whichever identifier the operator has handy. I expanded the doc comment on `LsDestinationItem` to capture this (bc5df11).

